PR #1, opened 2019-10-07
eip-0003: BIP-44 standard for Ergo Platform

gagarin55:
Before acceptance we need to discuss **coint_type** number and make PR to SLIP-0044 repository

### DIFF
--- a/eip-0003.md
+++ b/eip-0003.md
@@ -1,10 +1,10 @@
 BIP-44 standard for Ergo Platform
 =================================
 
-* Author: gagarin55
+* Author: gagarin55, Robert Kornacki
 * Status: Proposed
 * Created: 07-Oct-2019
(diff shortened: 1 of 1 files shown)


robkorn:
This may be a little bit of overly pedantic fun, but maybe we can make the *coin_type* number be `ergo` in ascii and summed?
`101 + 114 + 103 + 111 = 429`

`429` is still available based on the [list](https://github.com/satoshilabs/slips/blob/master/slip-0044.md), so it is doable.

gagarin55:
Why not? **429** might be good idea, only if they (SLIP-44) accept it

kushti:
Looks fine! If there are no more comments, lets merge?!

gagarin55:
Let's wait SLIP-0044 changes approval - https://github.com/satoshilabs/slips/pull/794

PR #2, opened 2019-10-07
eip-0004: Tokens standard

catena2w:


kushti:
@catena2w thanks for the submission! 

As was discussed on the forum (https://www.ergoforum.org/t/asset-issuance-standard-and-oracles/43/3), it is good to consider and standardize other applications of tokens than assets. For example, tokens are widely used in many application as account identifiers. A trivial case, tokens are being used to label oracle accounts, then possible types of oracles and feeds they are delivering can be standardized. 

Also, as most of the nodes are not storing historical data with easy access, it is worth to consider and standardize an issuance proof, consisting of a block header (maybe w. NiPoPoW proof) + issuance transaction along with a Merkle proof against the header.

In regards with the current writing, it could be improved in many places ("UDS", "Number of decimal place", unnecessary commas before "that" etc).

catena2w:
English is fixed.
I don't think it make sense to mix multiple approaches inside one EIP, let's create a separate one for oracles and others.
Regarding the issue info - it can be easily achieved via explorer right now, NiPoPoW proof looks like an overkill for such untrusted by default information.

eip-0004.md
```diff
@@ -0,0 +1,39 @@
...
+Storage rent component allows a miner to collect (or burn) all the tokens inside a box if it is expired, and there are not enough Ergs in the box to pay the storage rent fee.
+
+## Ergo tokens standard
+
+Though this is not required by the protocol, we propose the following structure for the box that issues a token:
+
+| Register        | Description                                     | Example                      |
+| --------------- |:-----------------------------------------------:| ----------------------------:|
+| R2              | Token id and amount pair                        | [("7d...09", 100000)]        |
+| R4              | Token verbose name (UTF-8 representation)       | "USD"                        |
```
aslesarenko: @catena2w It is worth mentioning that the strings are actually byte arrays or `Coll[Byte]` type of ErgoScript, with bytes obtained using the corresponding encoder (UTF-8 in this case).

catena2w: @aslesarenko fixed.
Also example of a real such transaction in mainnet was added

aslesarenko (approved):

kushti:
@catena2w let's rename the standard into "Assets standard" then. Issuance proof is not an overkill for a general token and needed, for example, to prove that a token claimed to be a singleton token was indeed issued in a proper way. 

Thus if you do not want to make more work now, it is better to rename the EIP into "Assets standard", aiming to have "Oracle Tokens Standard" and "Singleton Tokens Standard" (and more standards?) in future.

catena2w:
@kushti May be `Tokens issue standard `? 
Issue of singleton/oracle tokens is just a particular case of this eip (amount=1, decimals=0), while their usage for these purposes may be described in other eips.

kushti:
@catena2w It is better to name a document according to its contents. This document is not about token ISSUANCE in variety of known applications, rather, just assets. For example, it is not describing how to enrich oracle toked issuance data to have a possibility to build directories like https://oracles.wavesexplorer.com/

kushti (approved):

PR #4, opened 2020-03-10
EIP-0005: Contract Template

aslesarenko:


greenhat (commented):
LGTM. Please remove two image files(leftovers).

eip-0005.md
```diff
@@ -0,0 +1,139 @@
...
+be used as template parameters.
+
+Field Name     |  Format            | Example             | Description
+---------------|--------------------|---------------------|-----------
+`NameLength`   | `VLQ(UInt)`        | 5                   | Length of `Name` bytes 
+`Name`         | `Bytes`            | "delta"             | User readable parameter name (string bytes in UTF-8 encoding)
+`DescLength`   | `VLQ(UInt)`        | 14                  | Length of `Description` bytes 
+`Description`  | `Bytes`            | "spending delay"    | User readable parameter description (string bytes in UTF-8 encoding)
+`Type`         | `Type`             | Int                 | ErgoTree type serialized according to section (see _Type Serialization_ section of [ErgoTree Spec](https://ergoplatform.org/docs/ErgoTree.pdf)
+`Placeholder`  | `VLQ(UInt)`        | 0                   | Placeholder index in constants sequence
```
greenhat: @aslesarenko I see a problem with a parameter being eliminated from the contract on certain conditions. For example, if we have a paramater `a` of type `Int` and use it in the contract like `let c = b * a` it'll be eliminated by IR when `a == 1`.  
To mitigate this we can add something like "Valid values" field, but specifying it becomes a non-trivial task.

aslesarenko: Why it will be eliminated? You probably mean some optimizations and rewritings that take place in the current implementation?

greenhat: Yes, this is happening in the current implementation.

aslesarenko: This is a big mistake I have to confess. We should fix this in the 4.0 HF. ErgoTree stored in the blockchain have to be interpreted exactly as it is. This EIP may be useless for some contracts in the current version, but I suggest to keep it as simple as possible in favor of the future version after HF.

aslesarenko: The new ErgoTreeEvaluator interpret the trees directly and at least for all existing tests the result of new and the old interpreter are equal, which is good sign we will not have problems with switching the interpreter.

greenhat: I suggest we add 4.0 HF as a prerequisite/requirement to this EIP.

greenhat: I found another potential issue. In DEX sell order contract (ErgoScript) `tokenPrice` parameter is stored twice in the ergo tree constants. I suppose the reason is that it's used in two places and brought in via env. I assume that introducing a variable with its value from env in the contract code should eliminate the issue and leave only one constant. However, it seems like treating constants as contract parameters are very fragile and we need to introduce explicit ErgeTree parameters.

EDIT: Actually, it does not mess with treating constants as parameters to parse parameters from serialized ErgoTree. But I'd refrain from "changing" constant values to construct the new ErgoTree. 

https://github.com/ergoplatform/ergo-contracts/blob/5031e3f7dc05d981a03a8c6c280a16826f0c27fd/verified-contracts/src/main/scala/org/ergoplatform/contracts/DEXPartialFilling.scala#L88-L134

```scala
  def sellerContract(params: DexSellerContractParameters): ErgoContract = {
    // both value cannot be 1, otherwise compiler reduces b = a * 1 to b = a,
    // eliminating them from the expression (and from ErgoTree.constants)
    require(params.dexFeePerToken > 1, "dexFeePerToken should be > 1")
    require(params.tokenPrice > 1, "tokenPrice should be > 1")

    val sellerContractEnv =
      Map(
        "sellerPk"       -> CSigmaProp(params.sellerPk),
        "tokenId"        -> params.tokenId.toColl,
        "tokenPrice"     -> params.tokenPrice,
        "dexFeePerToken" -> params.dexFeePerToken
      )

    val sellerScript = s""" sellerPk || {

      val selfTokenAmount = SELF.tokens(0)._2

      val returnBox = OUTPUTS.filter { (b: Box) =>
        b.R4[Coll[Byte]].isDefined && b.R4[Coll[Byte]].get == SELF.id && b.propositionBytes == sellerPk.propBytes
      }(0)

      val foundNewOrderBoxes = OUTPUTS.filter { (b: Box) =>
        b.R4[Coll[Byte]].isDefined && b.R4[Coll[Byte]].get == SELF.id && b.propositionBytes == SELF.propositionBytes
      }

      (returnBox.value == selfTokenAmount * tokenPrice) || {
        foundNewOrderBoxes.size == 1 && {
          val newOrderBox = foundNewOrderBoxes(0)
          val newOrderTokenData = newOrderBox.tokens(0)
          val newOrderTokenAmount = newOrderTokenData._2
          val soldTokenAmount = selfTokenAmount - newOrderTokenAmount
          val minSoldTokenErgValue = soldTokenAmount * tokenPrice
          val expectedDexFee = dexFeePerToken * soldTokenAmount

          val newOrderTokenId = newOrderTokenData._1
          val tokenIdIsCorrect = newOrderTokenId == tokenId

          tokenIdIsCorrect && soldTokenAmount >= 1 && newOrderBox.value >= (SELF.value - minSoldTokenErgValue - expectedDexFee)
        }
      }

      }""".stripMargin

    ErgoScalaCompiler.contract(sellerContractEnv, sellerScript)
  }

```

aslesarenko: @greenhat There are different things which have to be corrected or/and made more flexible:
1) It is specific behavior of ErgoScript front-end which creates duplicate constants, one can always craft the ErgoTree directly as necessary, and we should fix ErgoScript compiler to generate more predictable ErgoTrees. Currently the code you write, and the ErgoTree you get may have different structure (while preserving the semantics). This is trade-off, optimizations versus predictability of the resulting template. We now favor templates a lot, and find more use cases for them, so optimizations seems like less important. 
2) In the new version of ErgoScript frontend, I think it is better to turn off the optimizations completely and instead provide the exact translation to ErgoTree, so that when writing ErgoScript you will know exactly which ErgoTree you will get.
3) Constant segregation logic is flexible enough to allow "partial" segregations. This however is not done by default in ErgoTreeSerializer.serialize method, which segregates all the constants. This is reasonable default, but in many cases you need to have more fine grained control of how the ErgoTree is serialized. 
4) ErgoTree Deserialization however fully supports partial constant segregation. So if you craft a partially segregated ErgoTree and put it into the transaction it will be deserialized correctly preserving exactly which constants are segregated. This was done with future parameters (i.e. this EIP) in mind.

So, on the protocol level this EIP is fully supported and doesn't require forking. But its usability may be complicated (depending on the concrete template). 
We can add 4.0 as usability prerequisite, but it is not strictly necessary and the EIP can be used with the current version.

greenhat: @aslesarenko I

I'm not sure ErgoScript frontend's update will resolve it. I believe the reason for two constants is two `LongConstant` IR nodes holding the same value which we cannot serialize as one constant in general.    
This brakes "## Conversion to ErgoTree" section.

> We can add 4.0 as usability prerequisite, but it is not strictly necessary and the EIP can be used with the current version.

I believe that 4.0 is not a usability prerequisite, but a required one. Otherwise, we cannot guarantee that declared  
parameters will be in `ErgoTree.constants`.

aslesarenko: @greenhat Could you please give a simplified example of the problem?
Where "two LongConstant IR nodes holding the same value which we cannot serialize as one constant in general" come from and why you make this conclusion?

aslesarenko: Also which example you keep in mind for "This brakes "## Conversion to ErgoTree" section."?

greenhat: @aslesarenko The seller contract (see above) gets compiled into the following tree:

----------------------------------------------------------

BlockValue(Vector(ValDef(1,List(),ConstantNode(SigmaProp(ProveDlog(ECPoint(4c6e23,afbbc9,...))),SSigmaProp)), ValDef(2,List(),SelectField(ByIndex(MethodCall(Self$(167),SMethod(SBox,tokens,(SBox) => Coll[(Coll[SByte$],SLong$)],8,MethodIRInfo(Some(<function1>)),Some(OperationInfo(Some(PropertyCall$(219)),Secondary tokens,ArrayBuffer(ArgInfo(this,this instance))))),Vector(),Map()),IntConstant(0),None),2)), ValDef(3,List(),Filter(Outputs$(165),FuncValue(Vector((3,SBox)),BlockValue(Vector(ValDef(5,List(),ExtractRegisterAs(ValUse(3,SBox),R4,Option[Coll[SByte$]]))),BinAnd(BinAnd(OptionIsDefined(ValUse(5,Option[Coll[SByte$]])),EQ(ValUse(5,Option[Coll[SByte$]]).get,ExtractId(Self$(167)))),EQ(ExtractScriptBytes(ValUse(3,SBox)),ExtractScriptBytes(Self$(167))))))))),SigmaOr(List(ValUse(1,SSigmaProp), BoolToSigmaProp(BinOr(EQ(ExtractAmount(ByIndex(Filter(Outputs$(165),FuncValue(Vector((4,SBox)),BlockValue(Vector(ValDef(6,List(),ExtractRegisterAs(ValUse(4,SBox),R4,Option[Coll[SByte$]]))),BinAnd(BinAnd(OptionIsDefined(ValUse(6,Option[Coll[SByte$]])),EQ(ValUse(6,Option[Coll[SByte$]]).get,ExtractId(Self$(167)))),EQ(ExtractScriptBytes(ValUse(4,SBox)),SigmaPropBytes(ValUse(1,SSigmaProp))))))),IntConstant(0),None)),Multiply(ValUse(2,SLong$),  **LongConstant(8417271212803245197)** )),BinAnd(EQ(SizeOf(ValUse(3,Coll[SBox])),IntConstant(1)),BlockValue(Vector(ValDef(4,List(),ByIndex(ValUse(3,Coll[SBox]),IntConstant(0),None)), ValDef(5,List(),ByIndex(MethodCall(ValUse(4,SBox),SMethod(SBox,tokens,(SBox) => Coll[(Coll[SByte$],SLong$)],8,MethodIRInfo(Some(<function1>)),Some(OperationInfo(Some(PropertyCall$(219)),Secondary tokens,ArrayBuffer(ArgInfo(this,this instance))))),Vector(),Map()),IntConstant(0),None)), ValDef(6,List(),Minus(ValUse(2,SLong$), SelectField(ValUse(5,(Coll[SByte$],SLong$)),2)))),BinAnd(BinAnd(EQ(SelectField(ValUse(5,(Coll[SByte$],SLong$)),1),ConstantNode(Coll(-88,-61,127,127,-128,1,0,-1,-128,10,-1,0,112,-87,-128,1,-120,-128,26,-128,127,73,-3,3,0,-37,-128,-114,1,0,-48,-61),Coll[SByte$])),GE(ValUse(6,SLong$),LongConstant(1))),GE(ExtractAmount(ValUse(4,SBox)),Minus(Minus(ExtractAmount(Self$(167)), Multiply(ValUse(6,SLong$),  **LongConstant(8417271212803245197)** )), Multiply(LongConstant(4376297012573121382), ValUse(6,SLong$))))))))))))

----------------------------------------------------------------------

See `LongConstant(8417271212803245197)` (`tokenPrice`) is in two places. Serializer extracts it as two constants in `ErgoTree`. It means that one contract parameter `tokenPrice` is "stored" in two constants. In order to make a new ErgoTree with different `tokenPrice` value, one has to change two constants. It invalidates the method described in "## Conversion to ErgoTree" section.

aslesarenko: Ah, that is what I mean by ErgoScript compiler should be fixed and allow precise control of the resulting ErgoTree. 

Having such constant duplication in ErgoTrees is the bad idea in a first place. 
You can avoid this by using `val tokenPrice = _tokenPrice` declarations in a script, which will embed the constant once in a single ValDef node. (I hope it should work in the current ErgoScript compiler).

aslesarenko: In addition, a smarter compiler can generate many ConstantPlaceholders referencing the same constant. This is also valid ErgoTree supported by the current deserializer and interpreter.
So one-to-one mapping between parameter and a _subset_ of constants should be enough to this eip. For EIP6 this requirement is more strict, they should map strictly one-to-one.

greenhat: > In addition, a smarter compiler can generate many ConstantPlaceholders referencing the same constant. This is also valid ErgoTree supported by the current deserializer and interpreter.
> So one-to-one mapping between parameter and a _subset_ of constants should be enough to this eip. 

To achieve it we need an updated compiler. That's why I wrote that a new compiler ("4.0") is a prerequisite for this EIP. I don't see how this EIP can be implemented with the current compiler.

aslesarenko: I suggested a workaround for constant duplication above, no?
Before EIP is implemented it should be accepted, that is what we discussing right now. 
And 4.0 is mostly about hard fork of the protocol, which is a lot of things to do even without front-end part.

greenhat: I accept this EIP. However, we have two issues:
1. Contract parameter is missing from constants (in `c = a * b` when `a = 1` it gets reduced by the compiler).
2. One parameter is stored in two constants.

The first issue can be resolved only with HF ("4.0"), right?  The second issue can be resolved either with HF ("4.0") for any contract or by specially crafting ErgoScript (introducing variable). Thus, I see that this EIP can be implemented only after HF, but the header says "Forking: not needed ". I don't think this is correct. 

EDIT: I'm assuming that this EIP should be applicable to any ErgoTree(contract).

aslesarenko: @greenhat There is a number of points which should be kept in mind: 
1) This EIP has nothing to do with frontend in general and ErgoScript in particular. There maybe many different frontends and compilers generating ErgoTrees (you are kind of doing the second as Scala macros). So the problems of anyone of them must not influence the design choices of the more foundational standard. ErgoTree is all we need to consider, and this standard doesn't assume any changes in ErgoTree, that is why "Forking: not needed". This is in contrast to EIP6, where ErgoTree is indeed need to be changed and "HF is required".
2) The ErgoTree format allow some flexibility. The format itself doesn't define all the constraints on the scripts. In other words, the fact that you can deserialize ErgoTree doesn't mean that ErgoTree is correct. It may have type errors, semantic errors, or even logical errors which make it fail in any context. All this additional constrains are defined as part of the ErgoTree specification, if they are not, then they should. Because all this constraints are parts of the protocol.
3) In a similar way, when it comes to defining templates we impose additional constraints on the consistency and correctness of the ErgoTree contents (since ErgoTree is a part of a template, not the over way around). This is necessary to make usage of templates: 1) unambiguous (deterministic); 2) simple for implementations in applications and tooling
Any additional over-complexity will be multiplied in the applications and tools, which I would like to avoid.

Keeping this in mind, for the two issues mentioned above. The issue 1) is the particular drawback of the particular frontend, should be fixed in the frontend. The issue 2) have root cause in the frontend also, but in addition, thinking about supporting it leads me directly to the over-complexity argument above.

greenhat: @aslesarenko Got it. Thanks for the detailed explanation. I support this EIP to be a standard.

aslesarenko: @greenhat If all the issues addressed, please consider approving

greenhat (approved):

eip-0005.md
```diff
@@ -0,0 +1,170 @@
...
+
+The ErgoScala compiler macros maps the parameters of the Scala method to the segregated
+constants, however those constants are mixed with other constants in the contract
+body (which are not mapped to template parameters).
+
+By using annotations the ErgoScala macros now can map parameters to constant positions (in
+ErgoTree.constants array) and use this mapping both in compilation of ErgoTree and
+generation of the template bytes. In particular, when creating ErgoTree bytes, the ErgoScala macros can do
+partial constant segregation, i.e. segregate only those constants from the expression tree
+which are mapped to the parameters leaving others in the tree. Thus created contract template will also satisfy EIP6
```
kushti: EIP7 perhaps?

eip-0005.md
```diff
@@ -0,0 +1,170 @@
...
+
+In addition to the binary format, the contract templates can be represented using JSON.
+It may serve better for some applications. Below we define a standard JSON representation
+of contract templates.
+
+```json
+{
+  "name": "rewardOutputScript",
+  "description": "holds mining rewards",
+  "constants": "0204a00b08cd0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
```
kushti: How constants are bound to placeholders? Is providing all of them mandatory?

eip-0005.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Parameter Serialization Format
+
+Parameters specify additional meta information for the contract template. The number of
+parameters can be less then equal the number of segregated constants. This is because some
+constant values may not be changed without breaking contract logic and hence they cannot
+be used as template parameters.
+
+Field Name     |  Format            | Example             | Description
+---------------|--------------------|---------------------|-----------
```
kushti, line 130: Which fields are mandatory and which are not?

kushti (changes requested):

eip-0005.md
aslesarenko: fixed

eip-0005.md
aslesarenko: described

eip-0005.md
aslesarenko: Described

kushti (approved):

PR #5, opened 2020-04-13
EIP-0007: Well-known Contracts Format

aslesarenko:


eip-0006.md
```diff
@@ -0,0 +1,163 @@
...
+
+Slot           | Format      | #bytes   | Description
+---------------|-------------|----------|-----
+_header_       | VLQ(UInt)   | [1, *]   | The header with `Bit 5` set to 1
+...            |             |          | other slots of ErgoTree
+root           | VLQ(UInt)   | [1, *]   | `contractCode` of a registered well-known contract
+
+To deserialize the ErgoTree with `WellKnownContractBit == 1` every compliant protocol
+implementation should keep a mapping between contract codes and contract templates. When
+the `contractCode` is deserialized it can be used to lookup the contract template in the
```
greenhat, line 145: Assuming that known contracts are hardcoded in the node source, how long does the author of the new known contract have to wait before they start putting it on the blockchain? Until the node is updated to recognize this new contract code it'll just reject the tx. So to safely use the new known contract one has to wait until the majority of nodes are having this contract hardcoded. We need some kind of "report" that shows the "adoption" of the new known contract by the nodes.

eip-0006.md
```diff
@@ -0,0 +1,163 @@
...
+`contractCode`   | C + 1              | The next code of the contract
+`prevContractId` | `lastContractId`   | The hash of the previous registered contract 
+`contractBytes`  | `Base16` or `File` | Contract Template bytes according to [EIP-0005](eip-0005.md) 
+`contractSig`    | `Base16` or `File` | The `signatureBytes` as Base16
+
+### Well-known Contracts List
+
+Each well-known contract must be added here in a sub-section and described according to
+the [well-known contract format](#well-known-contract-format). Each contract refers to the
+previous contract using its id thus forming a chain of contract registrations. The
```
greenhat, line 93: Why store the contracts in the text file here? I believe `ergo-contracts` or similarly organized repo would be a better fit. We can treat contracts as code, using PRs, formal verification and tests. To make a decision to include a new contract to the list one needs to understand the contract first. I believe formal verification, tests, etc. would help a lot.

greenhat (commented):
Looks good! I have two questions. :)

eip-0006.md
robkorn: This is one of the big questions I also have for this scheme. Where are the contracts specifically held? Are they hardcoded in the source? I don't think it's a great idea as it's very slow to add to, will cause validation issues if a Validator does not update their node in time and attempts to process a tx that has a well known contract that they don't have, and it's just not that clean.

Though it's much more complicated to implement, I think some sort of DHT is likely to be a better solution. All nodes (or just validating nodes) store a DHT which has the key map to the contract. Appending to the DHT can be dynamic via node API, and we simply implement a rule that only allow nodes to accept appending to the DHT by having a valid well-know contract that includes a 2-of-6 threshold sig from the foundation. As such, any node that is mining will have access to the DHT and has the guarantee that they will have access to the contract even if they happen to miss out on updating their node by a day or two. Encoding more or less complex rules is possible, and also adding rules for eventually removing contracts from the well-known list if for example they are found to be broken (maybe with a 5-of-6 threshold sig for example).

There's a lot more flexibility with this kind of setup, and it allows for projects to get their contract added dynamically as soon as they acquire the signature from the foundation.

aslesarenko:
@greenhat @robkorn Thanks for suggestions.  Originally I also thought that the activation of new contracts should be somehow part of the protocol. 
However from standardization point of view I see here two different aspects, one is data formats and another is activation protocol. 

Having separate and more focused EIPs will make them easier to support by tooling and apps. 
So this EIP6 should be renamed to EIP-0006: Well-known Contracts _Format_ and focus solely on data format part of the protocol.

The format should be self-sufficient, so that all software (including nodes, tools, apps etc) will be able to validate the consistency of the contracts they deal with and unambiguously interpret the content of the data. And the tools are not created equal, some of them don't care about activation (such as contract compilation/preparation tools).

The question of how to store the WKCs on the network node is completely their implementation choice. However, I agree, the question of whether a contract "can be used" or "not yet" should be guided by some activation protocol.
This can be a separate EIP7: Well-known Contracts Activation, which I feel have to be clarified before this EIP can be finalized and accepted.
So I would keep this PR unchanged and I'd rather switch to EIP7 and clarify it first.

aslesarenko:
I think the easiest and more principled way of storing, distributing and signing the WKCs is using the blockchain itself. Will investigate this ASAP then.

kushti (changes requested):
I definitely not in fond of this proposal at all as it is described.

It introduces unnecessary trust in centralized point of failure (ErgoFoundation, but the name of the entity does not matter even).  Please note that EF can cease to exist in future (as well as any other entity).   

Distribution and deployment of well-known contracts is also unclear (see comments by Robert and Denys).

Also please note that the "feeProposition" contract is not fixed by the protocol, rather, it is agreement between users and miners which could be changed at any time. 

I think well-known contracts should be done in a similar way, i.e. they should be a result of an agreement between users and miners, and the (transaction verification) core protocol must be neutral to the contents of the transactions included into the blockchain (concrete implementations can take into account well-known contracts for the sake of optimizations).

The very good and deeply thoughtful example here is Bitcoin: while the verification protocol is friendly to all the scripts, concrete memory pool implementations have stricter rules and filtering out non-standard transactions (e.g. non-P2PK/P2PKH/P2SH transactions), See isStandard() and isStandardTx() functions in the Bitcoin Core reference implementation.

aslesarenko:
Renamed the EIP to focus on Data Format part of the problem. 
Activation is thus out of scope and deserve it's own EIP.
Will edit the text to cover Format-only scope.

PR #6, opened 2020-04-21
EIP-0006: Informal Multi-Stage Contract Specification Format

robkorn:
After speaking with @aslesarenko in Discord we decided it would make sense to have this take the EIP-6 place.

### DIFF
--- a/eip-0006.md
+++ b/eip-0006.md
@@ -1,163 +1,259 @@
-# EIP-0006: Well-known Contracts 
+# EIP-0006: Informal Multi-Stage Contract Specification Format
 
-* Author: aslesarenko 
+* Author: Robert Kornacki
 * Status: Proposed
-* Created: 12-Apr-2020
+* Created: 21-Apr-2020
 * License: CC0
-* Forking: hard-fork needed 
+* Forking: No fork needed 
 
 ### Contents
 - [Background And Motivation](#background-and-motivation)
-- [Well-known Contract Format](#well-known-contract-format)
-- [Well-known Contracts List](#well-known-contracts-list)
-- [Implementation of The Standard](#implementation-of-the-standard)
-  - [Changes in Serialization](#changes-in-serialization)
-  - [Changes in Costing](#changes-in-costing)
+- [How Informal Multi-Stage Contract Specs Are Written](#How-Informal-Multi-Stage-Contract-Specs-Are-Written)
+- [Alternative Formatting](#Alternative-formatting)
+- [Example NFT Auction Contract](#example-nft-auction-contract)
+- [Conclusion](#Conclusion)
 
 ### Background And Motivation
 
-Ergo is a blockchain designed for long-term survivability, resilience to attacks and
-security. To facilitate these goals the following parameters of the network should be
-optimized:
-- Blockchain storage size on full nodes
-- Speed of Block Candidate assembly from pooled transactions
-- transaction validation speed (txs/sec) 
-- minimal network traffic required for a node to stay connected and keep the pace of the
-blockchain growth (obtaining new blocks, broadcasting new candidates)
-- Node memory footprint 
+[Multi-stage contracts](https://link.springer.com/chapter/10.1007/978-3-030-31500-9_16) are a fascinating tool for implementing complex protocols on a UTXO-based system. They allow for great expressiveness and are generally approachable due to the fact that different phases of a protocol are cleanly separated into their own sub-contracts.
 
-One of the opportunities to improve the network is to exploit the _stable_ nature of many
-useful contract templates (Here _stability_ means the new versions of the contracts are
-not created very often, if created at all):
-1) Ergo uses many contracts as part of its core protocol like
-`feeProposition`, `rewardOutputScript`, `foundationScript` etc. The contracts are fixed
-and are not supposed to be changed. 
-2) Many contracts such as those used in ErgoMix, DEX, Crowd Funding, etc. may become
-widely reusable. Many applications may be created on top of them. Such contracts have a
-natural tendency to be stable to keep backward compatibility with applications that use
-them.
-3) Stability of contracts also plays well with security. Intuitively, it feels very
-unsafe to send coins to an unknown contract. On the other hand, it is generally safe to
-send coins to the _well-known_ contracts, which have well described properties,
-verified by community experts and signed by network developers (or Foundation board
-members).
+With that said, multi-stage contracts are extremely nascent as a model. This means that though they may be simpler when compared to alternatives, complexity is still going to be present when dealing with protocols that require numerous stages. Furthermore, we could benefit from conveying to others how a protocol is supposed to be implemented and/or work without writing a single line of code.
 
-However, despite the actual stability of many contracts, in the Ergo Protocol v1 their
-copies have to be saved in thousands of transactions and UTXO boxes which they protect.
-This significantly increases the blockchain size and UTXO memory footprint. It also have
-significant computational overhead due to parsing and cost limit control.
+As such it would be quite handy to have a format to informally define specifications for our multi-stage contracts that is understandable, easily traversable, and can be used as a guide for eventually writing both the on-chain and off-chain code.
 
-This Ergo Improvement Proposal standardizes the protocol extension to factor those
-well-known contracts out of serializable boxes, transactions, blocks and finally out of the
-blockchain storage. It also allows to reduce validation time of both new block candidates
-and broadcasted blocks.
 
-### Well-known Contract Format
-Each _well-known contract_ must be specified and registered as it is described in this section and then
-added to the [list in the following section](#well-known-contracts-list).
 
-To specify a new well-known contract one needs:
+### How Informal Multi-Stage Contract Specs Are Written
 
-1) Obtain a the last registered well-known contract from [the list of well-known
-contracts](#well-known-contracts-list).
+Let's look at the markdown-based schema in it's entirety and drill down how it works below.
 
-1) Obtain `lastContractId` and the current counter value `C` from the most recent contract
-registration record.
+```md
+# Contract Name
+Contract preamble
 
-1) Describe the contract template according to [EIP-0005](eip-0005.md) and obtain its
-serialized bytes `contractBytes`. We impose an additional requirement that the number of
-segregated constants of the contract template is equal to the number of parameters. This
-is important to allow the same template (with exactly the same bytes) to be reused across
-many different ErgoTree instances of well-known contracts.
+## ToC
+1. [Bootstrap Stage](<#Bootstrap-Stage>)
+2. [X Stage](<#X-Stage>)
 
-1) Sign the "lastContractId|contractBytes" bytes with `Foundation Signature` (which can be
-a threshold `2-out-of-6` or similar signature) and obtain `signatureBytes`.
+## Bootstrap Stage
+Preamble of how the contract is bootstrapped. 
 
-1) Add a new contract registration data as a sub-section of the [next
-section](#well-known-contracts-list) using the data described in th
+### List Of Paths
+- [Y Path](<#Y-Path>)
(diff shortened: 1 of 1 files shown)


aslesarenko:
@robkorn please rebase on top of master.

UPDATE: on top of eip5
UPDATE2: seems not rebase in necessary (I just changes the base branch)

eip-0006.md
```diff
@@ -0,0 +1,259 @@
...
+
+
+#### Leads To
+`Hyperlinked name of next stage` OR `Exit`
+
+---
+
+```
+
+As you can see above, each contract is made out of a number of **stages**. Each stage (except for the bootstrap stage) is a specific contract that the locks up tokens and data. Every stage has the ability to specify values which are hard-coded within the contract, and/or values which are stored in the registers.
```
aslesarenko:
```suggestion
As you can see above, each contract is made out of a number of **stages**. Each stage (except for the bootstrap stage) is a specific contract that locks up tokens and data. Every stage has the ability to specify values which are hard-coded within the contract, and/or values which are stored in the registers.
```

aslesarenko: @robkorn It is not always the case that "each stage is a specific contract". For example DEX doesn't fit nicely into a _simple_ FSM model.
If we look at the DEX from FSM point of view, then we actually have two FSMs (one for buyer and one for seller).
And Swap transaction is moving both of them one step forward, and so Swap doesn't belong to any of the FSMs.
If we describe DEX as two multi-stage contracts, where the Swap transaction should belong?

Keeping this in mind, either the template should be more flexible, or the limitation clarified.

FlowCards for example don't require such strict one-to-one mapping of the whole scenario to FSMs.
So, in some sense, FlowCards are conceptually more high-level construction, which can be decomposed to FSMs and their interaction.

eip-0006.md
```diff
@@ -0,0 +1,259 @@
...
+
+From each stage you have one or more **paths**, which are state transitions from the current stage to another (or to itself recursively). Each path specifies the required **inputs boxes**, **output boxes**, and **path conditions**. Path conditions are the logical checks which are encoded within the contract which must be met in order to allow for a state transition to occur and thus for said path to be taken.
+
+Finally, each path also **leads to** somewhere else. In other words, the state transition must move the tokens and data held under the current stage to another. Thus we specify the next stage that a path leads to or `Exit`. Exit signifies that this is an **exit path** meaning that the multi-stage contract is finished executing and the locked data/tokens can be arbitrarily spent (thus exit the multi-stage contract).
```
aslesarenko: @robkorn By "can be arbitrarily spent" do you mean they are protected by p2pk contract?

eip-0006.md
```diff
@@ -0,0 +1,259 @@
...
+
+
+## ToC
+1. [Bootstrap Stage](<#bootstrap-stage>)
+2. [Auction Stage](<#auction-stage>)
+
+
+## Bootstrap Stage
+
+This stage allows a user to initiate the auction. It requires that they provide their NFT and a box with Erg which represents the starting bid. There are no hard-coded values, registers, or path conditions because this is the bootstrap stage and thus there is no contract yet.
```
aslesarenko: @robkorn There is an abuse of terminology, you call the whole process "contract" which have this Bootstrap Stage. But here you say "there is no contract yet". In both cases the meaning of "contract" is not clear. 

As a suggestion, we can call the whole process "a scenario" and the spending conditions call "scripts".

eip-0006.md
robkorn: Right, these informal specs don't need to map onto a single person's viewpoint as a separate FSM. The DEX would be in a single multi-stage contract spec and you would have a separate "Buyer Bootstrap Stage" And "Seller Bootstrap Stage" which I mention multiple bootstrap phases are possible later in the document.

In regards to Swap transaction, in the buyer's box/stage you'd have a path that is labeled "Swap Path" and in the sellers box/stage you'd have a "Swap Path" as well. In these paths you clarify that as inputs (and in the conditions) the other person's box is required.

Essentially you have this cross-referencing going on where each actor is in a different stage, and their stages can intersect and be spent together into a new stage(s).

There doesn't need to be a single path for a given specification. You are describing the whole protocol, as you would in FlowCards, but in simple english and thus are stating all the different possible states and transitions which encompass the whole multi-stage contract.

eip-0006.md
robkorn: That is likely to be most common case, but I specifically meant that the current multi-stage contract is completed and so users can choose to spend the box into a p2pk or into a new contract if they so wish.

eip-0006.md
robkorn: Fair point, it is a bit confusing as I was trying to keep to current Ergo terminology.

I'm personally not a big fan of "scenario" because it is a bit vague, but I agree with "script". I think we can keep the whole "multi-stage contract" concept be labeled as a "contract", while  the code locking a box as a "script".

aslesarenko: I think we are raising very important foundational semantic and vocabulary questions here.
We are often quite sloppy when we use the same word "contract" to name what I think are three completely different things: 
1) _spending condition_ - is basically proposition in a box (this corresponds to the checks which guard the box spending)
2) _script_ - which is a combination of many _spending conditions_ (this corresponds to the lifecycle of a box with many spending paths, for example BuyOrder, SellOrder in DEX)
3) _scenario_ - is the whole process in which many contracts may interact (this is where the whole DEX belongs)

I don't like the word "contract" because it is highly overloaded and abused in Ergo. 
In Ethereum the word "contract" have a single meaning and the meaning is always clear. 
In Ergo we have the 3 above meanings, and I would rather avoid using "contract" entirely.

I see the value of this EIP is to establish such a semantically unambiguous vocabulary so that all other contracts descriptions can become more meaningful for an external readers and newcomers. 

While we call it "informal" we still should make it unambiguous on semantic level.
That is my expectation from this EIP.

eip-0006.md
aslesarenko: @robkorn It becomes too complicated IMO. 
The goal of this formalized spec is to make it clear to anyone, right? 
I think structuring the template after FSM is semantically problematic, because FSM is not the right model. It make sense to have stages though.

robkorn: The reason we are using markdown is to allow for linking in the ToC, the list of paths, and where each path leads to so you can follow from a single bootstrap stage -> all the way to the end of a protocol by following the paths. This way you don't get lost and you can understand exactly how one actor in the protocol must move forward.

I don't really understand how it's any more complicated to you than FlowCards for example. It may be verbose, but FlowCards are inherently a lot harder to understand because people need to understand the Ergo model of contracts completly, how state transitions work, the conditions which you write in a very short form, etc. Everything is laid out in English here, and it's considerably easier to understand because you have separate Bootstrap phases for each actor which you can follow along. FlowCards don't have clear entry-points for someone to start looking at where the protocol begins on the other hand.

This may end up just being difference in perspective/opinion, but FlowCards are genuinely much more complicated to me while this format describes what the protocol does in straight english with links to make it easier to traverse.

eip-0006.md
robkorn: Fair point. However we do live in the realm of blockchain, and straying away from using "contracts" because it may have been abused previously seems like a negative to the ecosystem as a whole as we're just making the barrier of entry higher for people to relate to concepts they currently understand.

Right, so with your definitions in this document currently:
- spending condition = path condition
- script = script
- scenario = contract

So a contract contains one or more scripts which contains one or more path conditions. Spending conditions can be a synonym for path conditions and used interchangeably. I think this is overall quite simple to explain to people.

I am very much against the use of "scenario" because the English connotation when explaining it to people sounds weird/very ad-hoc. People expect to write smart contacts, not scenarios. If we throw away the term "contract" and replace it with "scenario", marketing the capabilities of Ergo is going to also be extremely unintuitive and confusing (ex. if we keep "contracts" in marketing but use "scenarios" when being technical, people will not understand how the two are related and why they are called different things).

"Scenario" make a lot more sense to use for a specific instance of a contract (that is live or running in a test). So a scenario would be using Contract X, with Actor A at stage Y and Actor B at stage Z with the specifics like data held in the registers known.

So in short, a "contract" is a protocol encoded into multiple stages/scripts. This works with the whole "multi-stage contract" concept which Ergo relies on and should be approachable to someone entering into the ecosystem.

eip-0006.md
```diff
@@ -207,13 +207,13 @@ This path recurses the box back into the Auction stage with the new Ergs held in
 
 
 #### Outputs
-1. A new box locked under the Auction contract again which has the new bidder's public key in R4, the NFT, and the Ergs for the bid.
+1. A new box locked under the Auction stage again which has the new bidder's public key in R4, the NFT, and the Ergs for the bid.
```
aslesarenko: You mean "script" and not "stage", right?

eip-0006.md
```diff
@@ -1,4 +1,4 @@
-# EIP-0006: Informal Multi-Stage Contract Specification Format
+# EIP-0006: Informal Multi-Stage Specification Format
```
aslesarenko: @robkorn I think, we need to name the things. "Informal Multi-Stage Specification" of what? I agree that "scenario" is not good term here, but removing Contract makes the title less clear. With the absence of better name, let's keep the old title.

eip-0006.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Background And Motivation
+
+[Multi-stage contracts](https://link.springer.com/chapter/10.1007/978-3-030-31500-9_16) are a fascinating tool for implementing complex protocols on a UTXO-based system. They allow for great expressiveness and are generally approachable due to the fact that different phases of a protocol are cleanly separated into their own sub-contracts.
+
+With that said, multi-stage contracts are extremely nascent as a model. This means that though they may be simpler when compared to alternatives, complexity is still going to be present when dealing with protocols that require numerous stages. Furthermore, we could benefit from conveying to others how a protocol is supposed to be implemented and/or work without writing a single line of code.
+
+As such it would be quite handy to have a format to informally define specifications for our multi-stage contracts that is understandable, easily traversable, and can be used as a guide for eventually writing both the on-chain and off-chain code.
+
+
```
aslesarenko, line 67: @robkorn Let's add a "Terminology" section and define the basic terms like "path condition", "script", "hard-coded values" etc. (also contract maybe to disambiguate it at least)

eip-0006.md
aslesarenko: In addition let's change "Specification" to "Description" in the title to reflect what you described as "it were an essay rather than the clunky feeling of writing compiled code (or a formal spec)". 
And also change Format to Template.
So I think of title "EIP-0006: Informal Multi-Stage "something" Description Template".

aslesarenko (changes requested):

eip-0006.md
aslesarenko: Also, looking at the list of EIPs it makes me feel confused, because both the  Contract Template and the  Well-known Contracts EIPs use the "contract" term to mean "script".
While this EIP use Contract to mean the whole scenario.

eip-0006.md
robkorn: Yes, however a `stage` is tightly interlinked to the `script`underlying it and I personally think it's fine to overload locking a box under a specific `stage` or a specific `script` and both make sense.

eip-0006.md
robkorn: I think "Multi-Stage Protocols" works pretty well which is what scalahub suggested on Discord, and then indeed changing "contract" == "script" as you have suggested/used.

robkorn: I disagree with changing it to "Description", because these are informal specifications of multi-stage protocols. This isn't just a loose "description" which states how a protocol would work, but instead is a specification which follows a given semi-formal format that is written in plain english, thus being informal. The vast majority of specifications in software are informal/semi-formal, and none of them are labled as "descriptions", and I believe that changing to it only makes it more vague and less understandable.

PR #11, opened 2020-05-25
EIP-0008: New Transaction Validation with Just-In-Time Costing

aslesarenko:
UPD:
Completely revised version:
1) `rewardTx` and `chargeTx` removed thus senders are not penalized for invalid transactions
2) `collatInputs` are not required
3) JITC for reduction combined with AOT for Sigma Protocol proposition verification
4) Default cost limit selection is motivated and described (to make the switch transparent for existing apps)
5) Overall changes to Ergo Protocol v1 are minimized.


eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+  - [The Benefits of the JIT costing](#the-benefits-of-the-jit-costing) 
+- [JIT costing of ErgoTree](#jit-costing-of-ergotree) 
+- [Initial tests of transaction validity](#initial-tests-of-transaction-validity) 
+- [Changes In The Reference Implementation](#changes-in-the-reference-implementation) 
+
+### Background And Motivation
+
+To prevent DoS attacks every
+[script](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c31fa42adc8b752dcec2c440bb49474b51fad5f2/docs/spec/ergotree.pdf)
+in the Ergo blockchain have to be executed with the control of the execution costs (aka
```
kushti: have to => has to

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+collection types with at least the `map` operation leads to a much more complex
+implementation if we still want to have expressive enough language. The `fold` operation
+is even worse for AOT costing unless we strictly limit it to primitive aggregations.
+
+2) Poor approximation of cost. We don't know the exact execution trace of the script
+ahead of time, in particular we cannot decide which branch of `if` statement will be
+executed (since we don't know the result of condition expression). 
+As a result we have to estimate both branches and take the maximum. This inherent estimation
+error is multiplied when this `if` is part of the `map` operation. A similar approximation
+have to be done if the script accesses an element of a collection by an index. If the
```
kushti: have to => has to

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+funds for the gas required to execute a complex contract, than the amounts of funds you operate
+with (i.e. sending $10 and pay $100 for the gas doesn't have economic sense).
+
+### Ergo Transaction Validation With JIT Costing
+
+Here we describe an extension to the existing Ergo protocol, which allows to prevent execution
+of scripts beyond allotted limits. The extension doesn't require changes in the
+serialization format. Moreover, simple _transfer transactions_ are processed the same way so
+that 1) most existing applications (like exchanges) will not require any changes and 2)
+the new protocol doesn't incure additional overhead for simple transfer transactions.
```
kushti: incure => incur

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+    `change` output box is balancing the total amount in `collatInputs` with `feeLimit` (see
+    the figure). The `inSum` amount of `inputs` boxes should be balanced with the `outSum`
+    amount of `outputs` boxes. This balance should be independent of the balance of
+    `collatInputs`. This is necessary when the transaction is interpreted as the _charge
+    transaction_, in which case `inputs` are not spent and `outputs` are not created, so the
+    effect on the UTXO still remains balanced.
+
+    In an exceptional case of invalid transaction (due to the scripts returning false, or
+    `actCost > feeLimit`) the transaction will still be added to the block but will be
+    interpreted as a _charge transaction_ (see step 7 below).
```
kushti: This breaks UTXO semantics completely. A transaction in the UTXO model  is spending inputs fully.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+    - R4: a `Long` value of `costPrice` which is equal to the number of NanoERG to be
+    paid per unit of cost.
+    - R5: a `senderPk` public key belonging to the sender. The ownership is proved
+    during `tx` signing by spending at least one box protected by the `p2pk(senderPk)`
+    contract.
+    - R6: an interval of blockchain heights in which the transaction is valid, a pair
+    of two Int values.
+ 
+3) The transaction sender needs to convince miners that he has enough coins to pay for
+transaction fees up to `feeLimit`. He can do it by showing enough boxes protected by
```
kushti: no need to convince if there's fee output already

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+    Note, when the charge transaction is validated by the other miners, they first
+    validate it as a normal (valid) transaction `tx` of the block following steps 5) - 6).
+    The other miners have to confirm the INvalidity of `tx`, after which they also
+    interpret it as a charge transaction.
+
+8) After the block candidate is full, the miner uses the `validTxs` list to create a
+`rewardTx` and append it to the block (see the figure). The `rewardTx` spends all `feeOut`
+boxes and creates 1) one output for the miner reward and 2) one output for each `senderPk`
+whose transaction `tx` had actual `actCost < costLimit`, thus the `feeLimit - actCost *
+costPrice` is sent to `senderPk` address (see the `refundBox` box in the figure).
```
kushti: Not friendly to SPV clients, now they need to download additional transactions.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+4) The context independence of scripts and hence the exact prediction of `actCost` is a
+static property of ErgoTree, it can be determined by analysing ErgoTree for which Context
+properties the script uses. This can be controlled and enforced by the sender. Moreover,
+it is possible to extend this property to a wider range of scripts by introducing a
+_validity interval_ of a transaction (see the `R6` register of the `feeOut` box). This is
+an interval of the blockchain heights during which the transaction can be processed. Thus,
+any scripts which run during validation can assume that the current height is within that
+interval, but do not rely on the precise value of the current height. This will make
+`actCost` exactly predictable even for scripts what have conditions on `HEIGHT` (like
+`HEIGHT > deadline`).
```
kushti: Why to do additional changes in the protocol for HEIGHT context-independence only ? In examples we have dependency on a spending transaction is more critical.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+representing different changes in UTXO. This choice is fully deterministic for all
+transactions and thus all validating nodes have a consensus about state changes.
+
+6) In a long term, JIT costing enables an economic incentive to improve performance of
+block validation including script evaluation. The miners are interested to put as much
+transactions in every block as possible to maximize fee payments.
+Considering the pool is full of transactions, the amount of transactions which can be put
+into a new block is limited by two parameters:
+      - Block validation time is limited by 2 minutes tick interval between blocks (which
+      BTW should be adjustable via voting). The time budget for block assembly and
```
kushti: There are no any works showing how block interval can be readjusted SECURELY.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+      validation itself is the fraction of this time (say 30 seconds), as the block must
```
kushti: Likely a couple of seconds, see GKL15, SPV mining issues in Bitcoin etc.

eip-0008.md
kushti: Also please note that "false" could mean invalid signature (unlike Ethereum where signature validation done separately). Unlikely including inputs with invalid signatures could be done securely.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+
+An algorithm for execution of ErgoTree with JIT cost estimation is described in [ErgoTree
+Specification](https://ergoplatform.org/docs/ergotree.pdf).
+
+### Initial tests of transaction validity
+
+This conditions should be checked for every well-formed transaction.
+
+- All checks performed in Protocol v1
+- INPUTS contain at least one `p2pk(senderPk)` box
```
kushti: This throws away mixers and likely other applications

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+- current height is within `feeOut.R6` interval
+- the `inputs` total amount and the `outputs` total amount are balanced
+- the `sum(collatInputs) + change == feeLimit`
```
kushti: why change is here?

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+An algorithm for execution of ErgoTree with JIT cost estimation is described in [ErgoTree
+Specification](https://ergoplatform.org/docs/ergotree.pdf).
+
+### Initial tests of transaction validity
+
+This conditions should be checked for every well-formed transaction.
+
+- All checks performed in Protocol v1
+- INPUTS contain at least one `p2pk(senderPk)` box
+- current height is within `feeOut.R6` interval
```
kushti: Please note that the fee box is not a part of the protocol.

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+- the `inputs` total amount and the `outputs` total amount are balanced
+- the `sum(collatInputs) + change == feeLimit` 
+
+### Changes In The Reference Implementation
+
+- rewrite transaction
+- cost model is required for initial checks (the cost contributing to MaxBlockLimit)
+- implement voting for the time interval between blocks
```
kushti: See above

kushti (changes requested):

kushti:
@aslesarenko I would like to propose to not to break UTXO semantics, bringing fee model (apart from storage rent component) into the protocol, vote for block intervals (which confronts with NiPoPoWs and not analyzed in a rigor way) etc. Also please note that not just miners bear validation costs but all the full-nodes propagating transactions. And current applications should stay. Thus let's discuss how to minimize changes, not to make a lot of reckless ones.

eip-0008.md
aslesarenko: @kushti  what is GKL15?

aslesarenko:
> Thus let's discuss how to minimize changes, not to make a lot of reckless ones.

Yes, this is the goal of doing "proposal first": discussion and clarification. To streamline the process, lets follow the following simple rules: 
1) leave the comments "in context", related to specific lines of the text (avoid general comments at the end)
2) write separate issues/questions as separate comments, so they can be discussed and resolved independently
3) if there is a claim that something is broken, please give the reasoning, why and how, so the claim is clear and can be addressed.
4) avoid raising cosmetic issues and focus on semantics first.

After collecting and clarifying the issues/questions in this way I can create the next version.

eip-0008.md
aslesarenko: fixed

eip-0008.md
aslesarenko: fixed

eip-0008.md
aslesarenko: fixed

eip-0008.md
aslesarenko: DataInputs also break the original UTXO semantics. Which UTXO semantics you mean?

> lso please note that "false" could mean invalid signature (unlike Ethereum where signature validation done separately).

I don't get the point. What is the problem exactly? (if you mean it is a problem)

eip-0008.md
kushti: https://www.iacr.org/archive/eurocrypt2015/90560288/90560288.pdf , see security assumptions on the round length (in comparison with expected block interval)

eip-0008.md
aslesarenko: the key point here is 
"_He can do it by showing enough boxes _protected by
`p2pk(senderPk)` script_ corresponding to the public key stored in `feeOut.R5` register._"

The idea is that, if there is at least one p2pk input, it should be proved to make the transaction valid. And this proof will be checked, thus we can be sure that sender have enough funds (unspent boxes) to cover fees. And only such boxes can be used to cover feeLimit. The feeOut.R5 register is necessary in case many p2pk addresses are used as inputs and we should know which one can be used for covering fees.

eip-0008.md
aslesarenko: This JIT costing is only applied to _not Well-Known Contracts_. So most of the contracts supporting basic infrastructure can be registered as Well-known.

eip-0008.md
aslesarenko: This part of JIT costing depends on the concrete implementation of miner fees. 
They are both "a part" or "not a part" of the protocol  (whichever the definition of the protocol).
If some other fee protocol will be developed in future (if any), it should keep JIT costing in mind. 
I consider JIT costing is more important than some abstract fee protocol flexibility. 
Even if the new version of fees protocol will be added, it can always be made compatible with the needs of JIT costing.

eip-0008.md
aslesarenko: What?

eip-0008.md
aslesarenko: As describe in 3) of the protocol: "The
    `change` output box is balancing the total amount in `collatInputs` with `feeLimit` (see the figure)."

eip-0008.md
kushti: @aslesarenko data inputs are not spendable ever, and they are clearly separated from always to-be-spent inputs (which are spent if spending transaction is in the blockchain). 

> I don't get the point. What is the problem exactly? (if you mean it is a problem)

The most trivial attack is that malicious miners can always include pk("spender_key") boxes with invalid signatures (maybe, along with the spender-pk contracts) to get paid. could be fixed by complicating the protocol further (requiring valid signatures from pk("spender_key") boxes, but could be other vectors then you need to consider

eip-0008.md
kushti: why the fee can't be paid without all these complexities with inputs?

eip-0008.md
kushti: Then update text of this EIP accordingly, to make it clear that it depends on centrally signed contracts as well.

eip-0008.md
kushti: "All checks performed in Protocol v1" means protocol changes, right? On the definition of the protocol, see V() predicate in GKL 15.

eip-0008.md
kushti: Sorry comment wasn't saved. So readjusting block interval on the fly is not proven to be secure (and likely not secure), also, breaking security of NiPoPoWs.

eip-0008.md
robkorn: I would agree that this really does change the inherent model/semantics of UTXO system. I would err on the side of caution as something that major could incur serious problems down the line which we may not be able to see until they happen. (As we're changing from a properly formed tx accepted by a node **will** spend an input UTXO now to **may**)

eip-0008.md
```diff
@@ -0,0 +1,479 @@
...
+The cost calculation of ErgoTrees is implemented as it is defined in [the next
+section](#jit-costing-of-ergotree).
+
+On a high level, the new protocol for transaction creation, signing and validation is
+described as a series of steps. The anatomy of the transactions which are created is shown
+in the figure below.
+
+ 1) A transaction sender performs cost estimation as part of transaction signing.
+Transaction signing involves: 1) execution of the scripts for all the input boxes and 2)
+execution of sigma protocols to generate proofs. This happens using the current context of
```
robkorn: > 1) A transaction sender performs cost estimation as part of transaction signing.
> Transaction signing involves: 1) execution of the scripts for all the input boxes and 2)
> execution of sigma protocols to generate proofs.


Doesn't this incur a really high barrier of entry for light clients? This would mean they all need to have support for executing the sigma protocols which is not trivial to have implemented and is actually quite a bit to ask for implementors. 

This would likely also stop more specialized light-clients who only have a few key contracts they support with various registers they change across txs to perform actions.

robkorn (commented):
JIT costing is likely to be prohibitive to light clients which isn't great.

eip-0008.md
robkorn: This also means that the majority of novel privacy protocols on top of Ergo in the future must be signed off by a centralized committee before they are "allowed" to be on-chain (not need `p2pk(senderPk)`. That's a pretty big blow to decentralization/usability + likely can be considered a complete breach of the "Ergo Vision" where the platform is supposed to be permisionless and open.

eip-0008.md
aslesarenko: Well, to avoid the attack you mentioned above. This is exactly the mechanism to avoid it.

eip-0008.md
aslesarenko: See the answer below. The attack you mention is prevented by those checks.

eip-0008.md
aslesarenko: The Well-known contracts should not be centrally signed. 
I got the concern about it and fully agree. 
There has never be an intention to make them centralized (I will be more careful in future).
Will change this part of EIP7 and remove mentioning the Foundation.
The storage and activation of new contracts can be done via soft-forks and new releases. 
We don't need any special storage for that.
This soft-forkability is already part of Ergo implementation, so can use for this specific case.

aslesarenko: JIT Costing doesn't depend on WKCs. However they will allow keep existing application unchanged. Also WKCs represent a trade-off.

eip-0008.md
aslesarenko: To be honest, supporting full sigma interpreter is not hard problem at all, the language is simple and the interpreter is straightforward. 
We have at least two options to have sigma on ALL platforms (Rust and Dart based).
We are doing Rust, and after some time this will not be an issue.
So, yes, light clients will be able to sign transaction in normal way and estimate costs.
I don't see this is a big deal.
And, even without JITC we still have the same problem, right?

PS:The hardest part of existing interpreter is actually AOT costing which is not necessary on the client side.

aslesarenko:
> JIT costing is likely to be prohibitive to light clients which isn't great.

Please see my response above.

eip-0008.md
polarker: > The most trivial attack is that malicious miners can always include pk("spender_key") boxes with invalid signatures (maybe, along with the spender-pk contracts) to get paid. could be fixed by complicating the protocol further (requiring valid signatures from pk("spender_key") boxes, but could be other vectors then you need to consider

On the other hand, if a transaction depends on the external environment (e.g. the hash of the previous block), It might fail only when miners packing the transaction. This might introduce some attack vectors? With JIT costing, such kind of transactions will cost gas fee always

kushti:
@aslesarenko @polarker @robkorn Let me put this EIP in another perspective.

In the first place, block validation (consensus-critical verification protocol) works in the same way regardless a costing procedure, if the procedure ensures that maximum time to validate a block (on commodity hardware) is defined by the cost limit per block, and also validation time is secure ( << expected block interval, so 1-2 seconds at most in case of Ergo). If costing procedure is correct (there are no mispriced instructions leading to spam attacks like Ethereum suffering from back in Autumn'2016), then there are no critical spam issues on this level, thanks to PoW making producing blocks a costly venture.  Thus we should look and fix problems on other levels, namely, propagation and block assembly. 

In Bitcoin, a node propagating transactions is doing following steps: 
  network transfer -> parsing -> UTXO fetching (from db) -> signature checks 

in the worst case, a failure happens during the last stage (signature validation). There were spam attacks against UTXO fetching stage (https://en.bitcoin.it/wiki/CVE-2013-2293). Currently, if transaction is not valid, a source peer got some penalty with a ban happening when penalty score reaching some threshold AFAIK.

Similarly, in AOT-Ergo:
  network transfer -> parsing -> UTXO fetching (from db) -> AOT -> interpretation -> signature checks 

in JIT-Ergo:
  network transfer -> parsing -> UTXO fetching (from db) -> interpretation -> signature checks 

However, in both cases the best spam strategy is the same: do a transaction with a cost about the limit but invalid signature. In regards with EIP-8, an attacker can use a heaviest WKC contract to avoid collateral boxes.

Thus for both AOT- and JIT-Ergo, a solution is about finding a proper fee and penalization mechanism for the p2p network. Maybe also with breaking transactions into different classes (like standard and non-standard transactions in Bitcoin), but only on the mempool level.

eip-0008.md
aslesarenko: No, it means currently active protocol. 
By this item I refer to all existing validation checks collectively. 
Other items are new.
Clarified this in the text.

eip-0008.md
```diff
@@ -0,0 +1,418 @@
...
+### Transaction Verification
+
+The verification is performed by _verifier_ nodes for each new transaction `tx` either
+selected from the pool or contained in a new block candidate broadcast over the network.
+
+1) A verifier determines the `costLimit` parameter of the transaction which should be less
+than `MaxTxCost`. The limit is computed as `costLimit = tx.feeOut.value /
+tx.feeOut.R4[Long]` if the register is not empty, otherwise `costLimit = DefaultTxCost`. 
+The value in the `R4` register is the NERG price of a unit of cost (`costPrice`).
+The `MaxTxCost` and `DefaultTxCost` parameters are described in the [section below](#default-transaction-cost).
```
polarker: There is no definition of `DefaultTxCost`

aslesarenko: thanks, fixed

eip-0008.md
```diff
@@ -0,0 +1,424 @@
...
+### Maximum Transaction Cost
+
+It is [observed in [2]](#references) that when a script execution requires nontrivial
+computation effort, practical attacks exist which either 1) waste miners’ computational
+resources or 2) lead miners to accept incorrect script results. This _verifier's dilemma_
+drives rational miners to be well-incentivized to accept unvalidated blockchains.
+The solution proposed (see [2], Lemma1) incentivizes correct execution of scripts by rational miners
+under the condition that miners only accept to verify the transactions that has cost limit
+bounded by `MaxTxCost = epsilon * MaxBlockCost / N` where `MaxBlockCost` is the limit on block
+verification costs, `N` is the number of transactions in a block and `epsilon` is a
```
polarker, line 173: Miners could not calculate `MaxTxCost` when assembly a candidate block because `N` is not known

eip-0008.md
```diff
@@ -0,0 +1,424 @@
...
+ 
+4) When `tx` has predictable `tx.accCost` cost (i.e. all input scripts are
+context-independent) then `tx` verification will never exceed `tx.accCost`, and thus
+setting `feeOut.value = tx.accCost * costPrice` guarantees the transaction is not rejected
+as over-limit.
+
+5) When `tx.accCost` is not predictable, which may happen when `tx` have at least one
+context dependent script (i.e. at least one script have conditions on `HEIGHT`,
+`CONTEXT.headers`, etc), then the sender can choose any value for `feeOut.value` and the
+`costLimit` will be selected by the verifier (see [details](#maximum-transaction-cost))
```
polarker: I think `costLimit` should also be able to be set by sender in this case. Only when a user does not set `costLimit`, the verifier will use `MaxTxCost`.

eip-0008.md
aslesarenko: `MaxTxCost` is fixed, same as MaxBlockCost, and can be adjusted via voting. Here `epsilon` is a variable which compensate for changes in N to preserve equation. 
Please see the paper for rationale behind limiting per-transaction cost.

eip-0008.md
aslesarenko: @polarker Yes, if the users want they can put a value in `tx.feeOut.R4`, thus taking control of the tx cost limit. The MaxTxCost is only used if `tx.feeOut.R4` is not present.

eip-0008.md
polarker: `N` is the number of transactions in a block according to your definition. Blocks could have different numbers of transactions, so `N` is dynamic and therefore `MaxTxCost`. That's why I don't get it when you say `MaxTxCost` is fixed.

polarker: `N` is known to`verifiers`, so it's not a problem for `verifers`. But `N` is unknown to block assemblies when building the block, so it will cause issues when preparing block candidates when txs are sorted according to reward efficiency. Let me try to give a concrete example in the following.

polarker: Let's assume that `MaxBlockCost = 100, epsilon = 1`. In mempool, there are 11 txs: `Tx(50), Tx(10) * 10` where the numbers inside the parenthesizes are tx costs. Let's also assume that the rewards per unit are the same for all these transactions. If the block assembler takes `Tx(50)` first, then there is no way for it to add other txs, because `MaxTxCost` will be lower than 50. The assembler could take the 10 transactions `Tx(10) * 10`, but this will make the assembly algorithm complicated in general. However, without `MaxTxCost` limit, the block assembler could simply take `Tx(50), Tx(10) * 50` which is invalid in your proposal. Attackers could exploit `MaxTxCost` to make the `MaxBlockCost` underused using such kind of ideas.

eip-0008.md
polarker: > the`costLimit` will be selected by the verifier

Then maybe this sentence should be improved a bit, as it gives me the impression that the `costLimit` is always unknown and set by the verifier.

eip-0008.md
aslesarenko: Epsilon is helper variable, which will not be used in practice. I brought it here in order to draw the parallels with the paper, but it also brings some confusion. 
Epsilon is purely theoretical parameter, which is here to compensate for changes in N. 
When N changes so does the epsilon, while other parameters of the equation are fixed.

eip-0008.md
aslesarenko: It is mentioned in other parts of the text, but yes, please propose your formulation.

eip-0008.md
polarker: In that case, what would be the suggested values for `MaxTxCost` and `MaxBlockCost`? These values should be discussed and included in the proposal in my opinion.

eip-0008.md
polarker: How about the following if I don't misunderstand the details:
> When `tx.accCost` is not predictable, which may happen when `tx` has at least one context-dependent script (i.e. at least one script have conditions on `HEIGHT`, `CONTEXT.headers`, etc), then the sender can either 1) choose values for `feeOut.value` and `costPrice` based on estimation, or 2) choose a value for `feeOut.value` and use `MaxTxCost` for `costPrice` (see [details](#maximum-transaction-cost))

aslesarenko: Fixed

eip-0008.md
aslesarenko: Probably yes, but they may also be voted, so I'm not sure we should put any specific value in the EIP.

eip-0008.md
```diff
@@ -0,0 +1,425 @@
...
+### Transaction Verification
+
+The verification is performed by _verifier_ nodes for each new transaction `tx` either
+selected from the pool or contained in a new block candidate broadcast over the network.
+
+1) A verifier determines the `costLimit` parameter of the transaction which should be less
+than `MaxTxCost`. The limit is computed as `costLimit = tx.feeOut.value /
+tx.feeOut.R4[Long]` if the register is not empty, otherwise `costLimit = MaxTxCost`. 
+The value in the `R4` register is the NERG price of a unit of cost (`costPrice`).
+The `MaxTxCost` parameter is described in the [section below](#maximum-transaction-cost).
```
kushti, line 87: Giving such semantics to R4 already breaks backwards compatibility for all the apps and standards using registers (mixer, EIP-4, DEX etc)

kushti (changes requested):

PR #19, opened 2020-07-06
 EIP-0006: Informal Multi-Stage Contract Specification Format

robkorn:
EIP-0006, seemed to be rebased back onto eip5 by mistake (and missed noticing). Nonetheless, the format is being currently used for protocols and is mature enough imo to be merged into master.

### DIFF
--- a/eip-0006.md
+++ b/eip-0006.md
@@ -1,163 +1,259 @@
-# EIP-0006: Well-known Contracts 
+# EIP-0006: Informal Multi-Stage Contract Specification Format
 
-* Author: aslesarenko 
+* Author: Robert Kornacki
 * Status: Proposed
-* Created: 12-Apr-2020
+* Created: 21-Apr-2020
 * License: CC0
-* Forking: hard-fork needed 
+* Forking: No fork needed 
 
 ### Contents
 - [Background And Motivation](#background-and-motivation)
-- [Well-known Contract Format](#well-known-contract-format)
-- [Well-known Contracts List](#well-known-contracts-list)
-- [Implementation of The Standard](#implementation-of-the-standard)
-  - [Changes in Serialization](#changes-in-serialization)
-  - [Changes in Costing](#changes-in-costing)
+- [How Informal Multi-Stage Contract Specs Are Written](#How-Informal-Multi-Stage-Contract-Specs-Are-Written)
+- [Alternative Formatting](#Alternative-formatting)
+- [Example NFT Auction Contract](#example-nft-auction-contract)
+- [Conclusion](#Conclusion)
 
 ### Background And Motivation
 
-Ergo is a blockchain designed for long-term survivability, resilience to attacks and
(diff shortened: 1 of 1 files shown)


kushti (approved):

eip-0006.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+This allows one to use lists within each given section, thereby allowing for much more room to go into detail about specifics. This can be useful for contracts which use several registers across several boxes in their spending conditions for example.
+
+With all of that said, let's now look at an example informal specification which uses the above scheme/format to define an NFT auction protocol.
+
+
+# Example NFT Auction Protocol
+This protocol allows a user to initiate an on-chain auction for a NFT/singleton token which accepts bids in Ergs. This informal specification is based off of the discussion/original design from [this Ergo Forum thread](https://www.ergoforum.org/t/auctions-on-ergo/122).
+
```
aslesarenko, line 228: @robkorn I think it is better to make this EIP self sufficient and include the example description here entirely. Please make the description more "up to the point", as this is more technical doc.

eip-0006.md
```diff
@@ -0,0 +1,324 @@
...
+## Stage ToC
+1. [Auction Stage](<#Stage-Auction>)
+
+
+## Action ToC
+1. [Bootstrap Auction](<#Action-Bootstrap-Auction>)
+2. [Bid On Auction](<#Action-Bid-On-Auction>)
+3. [Conclude Auction](<#Action-Conclude-Auction>)
```
aslesarenko, line 236: This structure of ToCs a bit confusing, can we have a single ToC section, with internal hierarchy of items if necessary?
Also, this may be subjective, but I found myself always confusing Action with Auction in the text. So maybe use another example for readability.

eip-0006.md
```diff
@@ -0,0 +1,324 @@
...
+#### Condition 1
+...
+#### Condition 2
+...
+
+```
+
+This allows one to use lists within each given section, thereby allowing for much more room to go into detail about specifics. This can be useful for contracts which use several registers across several boxes in their spending conditions for example.
+
+With all of that said, let's now look at an example informal specification which uses the above scheme/format to define an NFT auction protocol.
```
aslesarenko, line 215: Please define what NFT stands for?

aslesarenko (changes requested):

eip-0006.md
robkorn: An internal hierarchy cannot be used because actions often use several inputs at different stages at the same time. Thus there is no clear way to have a hierarchy, as will be obvious in the Oracle Pool specs we release publicly soon.

As such, the actions & the stages need to be separated into their own ToC.

eip-0006.md
robkorn: Good point, added definition

aslesarenko (approved):

PR #21, opened 2020-07-25
EIP-0001 Post-Implementation Edits

kushti:
This PR is about post-implementation edits of EIP-1.

### DIFF
--- a/eip-0001.md
+++ b/eip-0001.md
@@ -20,9 +20,9 @@ the dApp itself with handling forks, confirmation numbers, and so on.
 This Ergo Improvement Proposal focused on extending the wallet to be able to serve the needs of external applications by providing 
 a flexible scanning interface and the possibility for applications to register scans with the wallet to ensure that they are tracked. Scans that have successfully passed are considered to belong to the application.
 
-Each scan has a given scan ID, and each box found that matches said scan is tracked by the wallet and thus is associated with the scan ID. Among possible scans, there is a one default scan  
-implemented by the node wallet. It is called the SimplePayments scan. This tracks the user's owned P2PK boxes. Other scans are not directly implemented inside of 
-the wallet but can be added by a user or external application.
+Each scan has a given scan ID, and each box found that matches said scan is tracked by the wallet and thus is associated with the scan ID. Among possible scans, there are some pre-defined scans   
+implemented by the node wallet, to track wallet's public keys and also mining rewards. Other scans are not directly implemented inside of 
+the wallet but can be added by a user or an external application.
 
 
 Specification: Scanning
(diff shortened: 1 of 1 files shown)


greenhat (approved):

eip-0001.md
```diff
@@ -30,19 +33,22 @@ A new request to scan is initiated which registers said scan to be checked for a
...
 * `OR(predicate1, predicate2, ..., predicateN)` - if one of the children predicates is true
 
+*value* field in the predicates above is about encoded sigma value, *assetId* is about just 32-bytes long byte array.
 
-The following is an example of a predicate which states that the registered scan will search for boxes that contain a given asset and contains the provided bytes in R1.
+The following is an example of a predicate which states that the registered scan will search for boxes that contain a 
+given asset and also the provided bytes in R1 (*"0e24..."* in the example is a byte array which contains a script 
+represented in ErgoTree form).
 ```haskell
 AND(
```
aslesarenko, line 48: Why haskell and not Scala?

eip-0001.md
```diff
@@ -30,19 +33,22 @@ A new request to scan is initiated which registers said scan to be checked for a
 A predicate (function which returns a boolean value for a box) is required to register a scan.
 Predicates available are:
 
-* `CONTAINS(register, bytes)` returns true if certain register contains given bytes. If *register* is missed, R1 will be scanned
-* `EQUALS(register, bytes)`returns true if certain register contains only given bytes. If *register* is missed, R1 will be scanned
+* `CONTAINS(register, value)` returns true if certain register contains given value. If *register* argument is missed, R1 (script register) will be scanned
+* `EQUALS(register, value)`returns true if certain register contains only given value. If *register* argument is missed, R1 will be scanned
 * `OFTYPE(register, typeId)` - if certain register is of certain type
```
aslesarenko: What is typeId, is it Sigma type serialized bytes?

eip-0001.md
```diff
@@ -120,14 +120,29 @@ Specification: List Of All Proposed Endpoints
 - GET: `/scan/boxesUnspent/{scanId}` - List all boxes that have been tracked by the scan and are still unspent/part of the UTXO set
 
 
-Specification: Simple Wallet Scan
+Specification: List Of Implemented Endpoints
+--------------------------------------------
+
+- POST: `/scan/register` - 3.3.0
+- POST: `/scan/deregister` - 3.3.0
+- POST: `/scan/stopTracking/{scanId}/{boxId}` - 3.3.0
```
aslesarenko: It is already too late, but maybe for consistency don't pass POST parameter as part of URL?

eip-0001.md
```diff
@@ -120,14 +120,29 @@ Specification: List Of All Proposed Endpoints
...
+- GET: `/scan/listAll` - 3.3.0
+- GET: `/scan/boxesUnspent/{scanId}` - 3.3.0
+
+Objects and enpoints description can be found in openapi.yaml file in [Ergo protocol reference client repository](https://github.com/ergoplatform/ergo).
+
```
aslesarenko, line 157: Since the main purpose of EIP is to be used by app, the semantics of each API method should be fully specified here (in a self sufficient manner).

It is better to put precise specification of each method here (since the reference implementation already finalized).

After publication, this specification will serve as baseline truth both for apps and future releases.

eip-0001.md
kushti: Actually, not implemented, so removed

eip-0001.md
kushti: It is not Scala as well, but highlighting looks good

eip-0001.md
kushti: Actually, that's how it was implemented

eip-0001.md
kushti: I guess a link to openapi.yaml would be enough. It is provided

aslesarenko (approved):
LGTM

PR #23, opened 2020-08-10
EIP-0012 - dApp-Wallet Web Bridge

rooooooooob:
Technical spec for wallet-dApp communication.

### DIFF
--- a/eip-0012.md
+++ b/eip-0012.md
@@ -122,6 +122,20 @@ type Box = {|
 ```
 Additional registers have string ids "R4", "R5" to "R9".
 
+### BoxCandidate
+
+A candidate (no TX/box IDs) for an output in an `UnsignedTransaction`
+```
+type Box = {|
+  value: Value,
+  ergoTree: ErgoTree,
+  assets: TokenAmount[],
+  additionalRegisters: {| [string]: Constant |},
+  creationHeight: number,
+|};
+```
+Additional registers have string ids "R4", "R5" to "R9".
+
 ### BoxId
 
 Hex string of the box id.
@@ -227,13 +241,12 @@ Uses `sigma-rust` rep - Hex string of the id.
 An unsigned transaction. Uses a modified version of the transaction representation in `sigma-rust`.
 ```
 type Tx = {|
-  id: TxId,
   inputs: UnsignedInput[],
   dataInputs: DataInput[],
-  outputs: Box[],
+  outputs: BoxCandidate[],
 |};
 ```
-This differs from `Tx` in that that `inputs` would be `[UnsignedInput]`, however we haven't signed this tx yet and also the outputs are merely candidates.
+This differs from `Tx` in that that `inputs` would be `[UnsignedInput]`, there is no ID, and the boxes are just candidates.
 
 ### TxId
 
--- a/eip-0012.md
+++ b/eip-0012.md
@@ -1,0 +1,277 @@
+# EIP-0012: dApp-Wallet Web Bridge
+
+* Authors: rooooooooob, Robert Kornacki
+* Created: 10-Aug-2020
+* License: CC0
+* Forking: not needed
+
+## Description
+
+This document describes a communication bridge between cryptocurrency wallets and Ergo dApps which interact via javascript code injected into a web context. The communication is one-way from the dApp to the wallet.
+
+## Motivation
+
+Distributed apps (dApps) often require access to a user's wallet in order to function, and this is typically from a web context. In other cryptocurrencies, such as Ethereum, this is done via a library such as web3 that injects code into a web browser that the dApp can interact with to create transactions or obtain address data using a user's wallet. This should be done in a manner that maximizes privacy and an easy user experience.
+
+## API
+
+The proposed API is limited to just the minimum wallet <-> dApp communication needed rather than providing lots of utility functions (tx building, data conversions, etc) or node-access (for UTXO scanning). This is different compared to web3 as that functionality could be modular in a separate library and Ergo smart contracts don't need as much direct node access for basic dApp functionality compared to Ethereum.
+
+This API is accessibly from a javascript object `ergo` that is injected into the web context. This object contains just the `request_read_access()` method upon dApp start.
+
+### request_read_access() -> Promise\<bool>
+
+If the API was not already injected, requests access to the API, and if the user consents, injects the full API similar to EIP-1102 into the `ergo` object. Returns true if the API has been injected, or false if the user refused access.
+Does nothing and returns true if the API was already injected.
+
+The wallet can choose to maintain its own whitelist if desired but this is on the wallet side and not a part of this standard. It could possibly also expose a subset of its state as responses to the other calls but this is also out of scope. All following API methods are injected upon consent.
+
+### check_read_access() -> bool
+
+Returns true if the full API was injected and we have read access, and false otherwise. This is potentially useful to have a cleaner UX on the dApp side (e.g. display request read access button for optional functionality rather than always requesting it).
+
+### get_utxos(amount: Value = undefined, token_id: String = 'ERG', paginate: Paginate = 'undefined') -> [Box] | undefined | PaginateError
+
+Returns a list of all unspent boxes tracked by the wallet that it is capable of signing for.
+Takes an optional {amount} parameter, if not `undefined`, limits the returned UTXOs to {amount} of the token {token_id} to create it. if {amount} is not `undefined` and there is not sufficient balance of the specified token to reach {amount} then `undefined` is returned.
+If {paginate} is not `undefined` then results are returned in a paginated manner in chronological order of use on the blockchain. If {paginate} is out of range, `PaginateError` is returned.
+
+### get_balance(token_id: String = 'ERG') -> Value
+
+Returns available balance of {token_id} owned by the wallet. A useful and minimal convenience functionality, that while could be served by iterating the return of `get_utxos` and summing, is useful enough and could be most likely quickly returned by wallets without needing to do this calculation.
+
+### get_used_addresses(paginate: Paginate = 'undefined') -> [Address]
+
+Returns all used (exist in confirmed transactions) addresses of the user's wallet.
+If {paginate} is not `undefined` then results are returned in a paginated manner in chronological order of use on the blockchain. If {paginate} is out of range, `PaginateError` is returned.
+
+### get_unused_addresses() -> [Address]
+
+Returns unused addresses available that have not been used yet. The wallet would need to pre-generate addresses here within its restrictions (e.g. discover gap on indices for HD wallets). These can be used for receive addresses.
+
+### sign_tx(tx: Tx) -> Promise\<SignedTx> | TxSignError
+
+A signed tx is returned if the wallet can provide proofs for all inputs (P2PK ones). If not it produces an error. This should also have some kind of user pop-up which if rejected would produce an error as well. This way signing/keys are kept entirely in the wallet.
+
+### sign_tx_input(tx: Tx, index: number) -> Promise\<SignedInput> | TxSignError
+
+Lower level tx signing functionality that signs a single input if the wallet is able to, and returns an error if not. This can be useful for constructing transactions consisting of inputs from multiple parties, and should be general enough for future transaction signing use-cases. Likewise with `sign_tx` this should ask for user consent and can be rejected producing an error.
+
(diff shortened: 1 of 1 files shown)


eip-0012.md
```diff
@@ -0,0 +1,277 @@
...
+* License: CC0
+* Forking: not needed
+
+## Description
+
+This document describes a communication bridge between cryptocurrency wallets and Ergo dApps which interact via javascript code injected into a web context. The communication is one-way from the dApp to the wallet.
+
+## Motivation
+
+Distributed apps (dApps) often require access to a user's wallet in order to function, and this is typically from a web context. In other cryptocurrencies, such as Ethereum, this is done via a library such as web3 that injects code into a web browser that the dApp can interact with to create transactions or obtain address data using a user's wallet. This should be done in a manner that maximizes privacy and an easy user experience.
```
kushti: "and an easy user experience" => "and provides an easy user experience" ?

eip-0012.md
```diff
@@ -0,0 +1,277 @@
...
+
+## API
+
+The proposed API is limited to just the minimum wallet <-> dApp communication needed rather than providing lots of utility functions (tx building, data conversions, etc) or node-access (for UTXO scanning). This is different compared to web3 as that functionality could be modular in a separate library and Ergo smart contracts don't need as much direct node access for basic dApp functionality compared to Ethereum.
+
+This API is accessibly from a javascript object `ergo` that is injected into the web context. This object contains just the `request_read_access()` method upon dApp start.
```
kushti: accessibly => accessible

eip-0012.md
```diff
@@ -0,0 +1,277 @@
...
+
+Signs the generic data {data} encoded as a hex string using the private key associated with {addr} if possible. If the wallet does not own this address, an error is returned. This should ask for user consent and produce an error if rejected. The wallet should also implement a message signing protocol such as the proposed [EmIP-005](https://github.com/Emurgo/EmIPs/blob/5b00fce84f31eb763892186eb9c88739ec809333/specs/emip-005.md) in order to make this endpoint safer for use, and make it harder for the user to accidentally sign data such as transactions, blocks, etc. Returns the signed data as a hex-encoded string.
+
+TBD: Hash algorithm/format to use.
+
+### submit_tx(tx: SignedTx) -> TxId | TxSendError
+
+Uses the wallet to send the given tx and returns the TxId for the dApp to track when/if it is included as its own responsibility. This is technically not mandatory but any wallet that is using such a bridge should have this functionality. Wallets can additionally choose to rate-limit this call, and can return an error if they did not or could not send the transaction.
+
+### add_external_box(box_id: BoxId)
```
kushti: output parameter is missed

eip-0012.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+* ProofGeneration - Wallet could not sign the data (e.g. does not have the secret key associated with the address)
+* AddressNotPK - Address was not a P2PK address and thus had no SK associated with it.
+* UserDeclined - User declined to sign the data
+* InvalidFormat - If a wallet enforces data format requirements, this error signifies that the data did not conform to valid formats.
+
+### Value
+
+BigNum-like object. Represents a value which may or may not be ERG. Value is in the smallest possible unit (e.g. nanoErg for ERG). It can be either a `number` or a `string` using standard unsigned integer text representation.
```
kushti: Value is a positive but encoded as i64 number actually.

eip-0012.md
```diff
@@ -0,0 +1,277 @@
...
+```
+or the empty `{}` object for P2PK inputs.
+
+### ErgoTree
+
+Uses `sigma-rust` rep - Hex string representing serialized bytes of the tree using `SigmaSerialize`
+
+### Input
+
+Read-only input (e.g. oracle usage or one yet to be signed) that is not spendable. Uses `sigma-rust` JSON rep:
```
kushti: There's a difference between read-only input and ordinary input: while the former contains box id, the latter also may contain context extension.

kushti (changes requested):

eip-0012.md
greenhat: @kushti Actually it was a "raw" u64 initially, but now I made a newtype `BoxValue(u64)`

eip-0012.md
rooooooooob: There was originally none but perhaps this is more useful to return whether or not the wallet is tracking it.

eip-0012.md
rooooooooob: If it's positive somewhere (even if Long/i64 encoded), then it should be fine as-is here, right? I didn't look into how Scala handles `Long` deserialization from JSON for numbers too big to fit, but in sigma-rust the wrapped u64 should take in a string-encoded uint like "1053546432154" afaik. Please correct me if I'm wrong and this should be changed.

eip-0012.md
rooooooooob: Updated to split into UnsignedInput/SignedInput/DataInput

eip-0012.md
kushti: You mean whether it was tracked before the call?

eip-0012.md
kushti: Scala representation of Long is signed, so max value is 2^63-1.

eip-0012.md
rooooooooob: Whether it's tracked after the call regardless of whether it was tracked before. If it's tracked before and you call it, it returns true. If it was untracked before and it's tracked now it returns true. If it was untracked before and the wallet can't or won't track it, it returns false.

eip-0012.md
rooooooooob: For JSON deserialization in the full node is this description fine though? Wouldn't the max possible value for nanoergs be much smaller than this? I guess for other token types it could be relevant - in which case should we keep it as-is but specify that it cant' be more than a 64-bit signed number or what?

gagarin55:
I suggest you to specify data types in language independent way, not types from sigma-rust implementation

rooooooooob:
> I suggest you to specify data types in language independent way, not types from sigma-rust implementation

We do specify the types in language-independant JSON. This was intended to mirror what sigma-rust had at the time for ease of interopability, but they are specified at the bottom of the document.

rooooooooob:
I made some minor changes based on issues encountered during our initial implementation inside of Yoroi:

Type definitions are specified using flow types now instead of some
pseudo JS schema.

API errors are added to all API calls to handle things that can go wrong
for any call,e.g. internal error in the wallet, incorrect params
specified to API calls.

`get_change_address()` was added as an explicit API since it should be
easy to support and allows for an easier way during tx creation rather
than arbitrarily choosing an address from the used or unused API calls,
which aren't guaranteed to be non-empty.

`add_external_box()` was removed as we are still unsure how any wallets
would be able to implement it in any meaningful way. This could be
reversed and have most wallets treat it as a no-op.

We removed `BoxCandidate` in favor of just `Box` to make things simpler. It also allows interop with sigma-rust/full nodes for now.

It would be ideal to not reference sigma-rust in any way for things that are more specific than just the underlying javascript type for example:
```
Uses `sigma-rust` rep - Hex-encoded bytes for `SigmaSerialize` of a constant.
```
which specify things like encoding, length, etc.

eip-0012.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+### UnsignedInput
+
+Uses full-node JSON rep:
+```
+type UnsignedInput = {|
+  boxId: BoxId,
+  extension: ContextExtension,
+|};
```
SebastienGllmt, line 194: Need to add the full content of `Box` to the input here that way the light wallet can properly show the user the content of the input (and we can validate the content is right by making sure it matches the box id passed in)

rooooooooob:
`BoxCandidate` was restored as it was only removed to ease support with sigma-rust/for simplicity, but the additional information was never necessary. Now that the data model was updated in sigma-rust it makes more sense to revert that change to the original `BoxCandidate` model and remove tx IDs from the unsigned `Tx` which makes it potentially easier to construct by the dApp.

Full box contents are now included in `UnsignedTransaction` to help light wallets support P2S inputs + for all wallets in processing recently created outputs. Wallets can check that the content matches the ID so there should be no security issues with that.

eip-0012.md
```diff
@@ -0,0 +1,348 @@
...
+  index: number,
+|};
+```
+Additional registers have string ids "R4", "R5" to "R9".
+
+### BoxCandidate
+
+A candidate (no TX/box IDs) for an output in an `UnsignedTransaction`
+```
+type Box = {|
```
greenhat:
```suggestion
type BoxCandidate = {|
```

eip-0012.md
```diff
@@ -0,0 +1,348 @@
...
+
+The initial design for most of these types are meant to be the same as the JSON types in the Ergo full node, which matches up as of now with the `sigma-rust` representations, but these definitions are the definitive types.
+
+### Address
+
+String as the standard base58 encoding of an address.
+
+### Box
+
+Transaction output box. Format is the same as `sigma-rust`:
```
greenhat, line 110: There is no `Box` in sigma-rust. I suppose `ErgoBox` is meant here.

eip-0012.md
```diff
@@ -0,0 +1,348 @@
...
+  index: number,
+|};
+```
+Additional registers have string ids "R4", "R5" to "R9".
+
+### BoxCandidate
+
+A candidate (no TX/box IDs) for an output in an `UnsignedTransaction`
+```
+type BoxCandidate = {|
```
greenhat, line 129: Probably should mention that it's `ErgoBoxCandidate` in `sigma-rust`?

eip-0012.md
```diff
@@ -0,0 +1,348 @@
...
+```
+type ProverResult = {|
+    proofBytes: string,
+    extension: ContextExtension,
+|};
+```
+
+### SignedTx
+
+Uses `sigma-rust` rep for a transaction:
```
greenhat, line 222:
```suggestion
Uses `sigma-rust` `Transaction` rep for a transaction:
```

rooooooooob:
We're currently also developing a similar spec to this for Cardano [here](https://github.com/cardano-foundation/CIPs/pull/88). There has been a lot of discussion there, some of which is relevant to Ergo. Here are some things that we might want to cover here too:

- Handling multiple wallets installed (for Cardano we solve this by injecting into a namespaced object - for Ethereum they just tell you to only have one installed e.g. MetaMask)
- Versioning (i.e. if this spec gets updated and some API changes)
- Phishing prevention (this could be assumed to be a wallet implementation detail, e.g. showing some user-known secret image in connector popups to stop phishing sites from creating similar looking popups to a user's wallet to farm passwords or other info)

eip-0012.md
```diff
@@ -0,0 +1,348 @@
...
+
+## API
+
+The proposed API is limited to just the minimum wallet <-> dApp communication needed rather than providing lots of utility functions (tx building, data conversions, etc) or node-access (for UTXO scanning). This is different compared to web3 as that functionality could be modular in a separate library and Ergo smart contracts don't need as much direct node access for basic dApp functionality compared to Ethereum.
+
+This API is accessible from a javascript object `ergo` that is injected into the web context upon wallet consent. Before this just the free `ergo_request_read_access()` and `ergo_check_read_access()` functions are injected into the web context. An event handler can also be registered to the web context to detect wallet disconnects tagged as `"ergo_wallet_disconnected"`.
```
capt-nemo429, line 21: In order to avoid conflicts with multiple wallets, can we do a slight change on the access API? Maybe something like `ergo.{walletName}.requestReadAccess()` and `ergo.{walletName}.checkReadAccess()`

rooooooooob: Yeah that's what we ended up doing for CIP-0030 for Cardano which I based off of this. I'm not sure if it's fine to do such a breaking change at this point since this has been up for a long time although it's still a draft PR. I'll wait and see what the Ergo devs think about this. We also don't have any functionality for versioning either. I mentioned both of these in my comment above in August. There are additional events that can be of use for dApps as well that could be added.

oskin1: Agree with @capt-nemo429 . Explicitly distinguishing wallets in the dApp connector API is a crucial feature. Shouldn't be too problematic to migrate.

eip-0012.md
MrStahlfelge: I think the wallet should fetch the contents itself from its information instead. It is done this way with ErgoPay and is absolutely feasible.

MrStahlfelge (changes requested):

PR #26, opened 2021-01-19
EIP-004 - Extension to include asset types standard

anon-real:
standardizing asset types e.g., picture artwork NFT.

### DIFF
--- a/eip-0004.md
+++ b/eip-0004.md
@@ -44,7 +44,7 @@ The transaction that issues such a token was included in block 98288 and may be 
 
 This standard is an extension of [token standard](#ergo-tokens-standard):
 
-In the asset type standard, R7 is a required two-byte value encoded as Coll[Byte]. The first byte represents the asset category, for example, _NFT_ or _Share Tokens_. The second byte specifies the exact subcategory, for example, _Picture Artwork NFT_ or _ErgoMixer's Share Tokens_. The second byte can be omited so that the issuance specifies only the catagory and not the subcatagory, for example, _Share Tokens_.
+In the asset type standard, R7 is a required two-byte value encoded as Coll[Byte]. The first byte represents the asset category, for example, _NFT_ or _Share Tokens_. The second byte specifies the exact subcategory, for example, _Picture Artwork NFT_ or _ErgoMixer's Share Tokens_. The second byte can be omited so that the issuance only specifies the catagory and not the subcatagory, for example, _Share Tokens_.
 Also, the remaining R8 and R9 registers can be used by each individual asset type based on its needs.
 
 The standardization of various asset types can be found below:
--- a/eip-0004.md
+++ b/eip-0004.md
@@ -45,7 +45,7 @@ The transaction that issues such a token was included in block 98288 and may be 
 This standard is an extension of [token standard](#ergo-tokens-standard):
 
 In the asset type standard, R7 is a required two-byte value encoded as Coll[Byte]. The first byte represents the asset category, for example, _NFT_ or _Share Tokens_. The second byte specifies the exact subcategory, for example, _Picture Artwork NFT_ or _ErgoMixer's Share Tokens_. The second byte can be omited so that the issuance only specifies the catagory and not the subcatagory, for example, _Share Tokens_.
-Also, the remaining R8 and R9 registers can be used by each individual asset type based on its needs.
+Also, the remaining R8 and R9 registers can be used by each individual asset types based on their needs.
 
 The standardization of various asset types can be found below:
 | Asset type        | R7                                     | R8                      |R9                      |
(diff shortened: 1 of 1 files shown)


kushti (approved):

jasondavies (changes requested):
I think this is confusing: `SHA256 hash of the artwork (UTF-8 representation)`.

I suggest that the register contain the SHA256 hash of the artwork, i.e. 32 bytes encoded as a Coll[Byte].

One of the problematic things about the original EIP is that the number of decimal places (an integer) is encoded as a UTF-8 string, but I think it's too late to change this now.

greenhat (commented):
I agree with @jasondavies on artwork hash encoding. 
Besides that, I’m wondering if string type in R7 is that much better than a byte array ([0x01, 0x01] vs. “0101”) if we’re going to treat it like two numbers after decoding anyway.

anon-real:
> I think this is confusing: `SHA256 hash of the artwork (UTF-8 representation)`.
> 
> I suggest that the register contain the SHA256 hash of the artwork, i.e. 32 bytes encoded as a Coll[Byte].
> 
> One of the problematic things about the original EIP is that the number of decimal places (an integer) is encoded as a UTF-8 string, but I think it's too late to change this now.

Makes sense, fixed!

jasondavies:
~@greenhat I think the intention is for R7 to contain `[0x01, 0x01]` as it doesn't mention UTF-8.~

Edit: you're right, it looks like it's a UTF-8 string encoding of "0101".

jasondavies (approved):

anon-real:
> I agree with @jasondavies on artwork hash encoding.
> Besides that, I’m wondering if string type in R7 is that much better than a byte array ([0x01, 0x01] vs. “0101”) if we’re going to treat it like two numbers after decoding anyway.

I think it is better to leave R7 as is for two reasons:
* Since it is just representative of asset type, it may not be necessary for it to be treated as numbers after decoding (unlike "Number of decimals".)
* Several artworks are around with "0101" in R7. Not a big deal, but if it is possible to keep them in the standard it is better to do so.

eip-0004.md
```diff
@@ -39,3 +39,20 @@ Though this is not required by the protocol, we propose the following structure
...
+| R7              | Asset type (UTF-8 representation)                       | "0101"        | "0e0430313031"|
+
+Note that the starting "01" in the asset type above is representing that the asset is NFT and the final two characters are standardizing the NFT type e.g., Picture artwork NFT. Also, the remaining R8 and R9 registers can be used by each individual asset type based on its needs.
+
+The standardization of various NFT types can be found below:
+| Asset name        | R7                                     | R8                      |R9                      |
+| --------------- |:-----------------------------------------------:| ----------------------------:|----------------------------:|
+| Picture artwork NFT              | "0101"                        | SHA256 hash of the artwork    | Optional - link to the artwork (UTF-8 representation) |
+
+The above registers (R7-R9) are also encoded as Coll[Byte] type.
```
aslesarenko, line 56: In general LGTM. A couple of questions:
1) We have a trade-off here like compactness vs flexibility, with strings being more flexible, but also less space efficient. Should we just use bytes directly for type encodings?
2) I suggest to specify more explicitly the internal structure of R7 and also give an example of non-NFT asset, how it can be encoded. For example, should I interpret and R7 with leading "01" string prefix as NFT. For example if R7 == "01", should I interpret it as NFT in which NFT-type is not specified? Or does it represent some non-NFT asset? Or should it be considered invalid asset issuance?
In order to answer these questions unambiguously, the parsing rules of the R7 register should be defined (regardless of how we encode it: string or bytes, see point 1) above).
At the same time the rules can be open for extension, for example saying something like "after the prefix is parsed, the rest of the R7 is parsed depending on the value of the prefix, which can be defined in further extensions of the standard".

aslesarenko (changes requested):

eip-0004.md
anon-real: @aslesarenko 
1. I agree on the space trade-off but maybe not that much of a difference and since there are some artworks already encoded like that better to keep it like this. However, if you think it is actually a big difference for space, then I'll change it.
2. Good point, I'll be more specific. as an example, starting with "02" can be for share tokens; then the leading "01" can be for the mixer for example, e.g., "0201", mixer's share tokens. Another example: "0102", Erg/USD oracle NFT. Then, if an issuance contains only the first two leading characters, it can be valid but not exactly specified e.g., "02" share tokens, "01" NFT. If you think this logic is sound, then I will add its description to the standard.

aslesarenko: 1) Just realized that bytes-based format has additional advantage. Such type codes can be parsed directly from contracts (whereas String and UTF-8 is not supported by ErgoScript). 
So, even if for some reason we will really need to have UTF-8 string as part of R7, then it still can be specified as part of prefix-dependent parser (i.e. when prefix byte = 02 and the remaining bytes are expected to be UTF-8 string)
Having String in R7 makes sense if there are cases when the type code contains alphabetical characters. However even in this case I would still prefer to have byte codes fixed in the standard and stored in the blockchain as much as possible.
2) yes, this logic makes sense, please describe the parsing rules so that interpretation is unambiguous

anon-real: Addressed all your comments. Added more description for the meaning/parsing of R7 and also it is now encoded as byte codes.

aslesarenko (approved):

PR #30, opened 2021-05-19
EIP-0017 - proxy contracts

anon-real:
Outsourcing transaction generation to an external service/dApp can be useful or needed in various circumstances. For example, avoiding wallet limitations to generate any transaction on behalf of the user - [Ergo Assembler](https://github.com/anon-real/ergo-assembler) is designed for this purpose. Another example is to scale dApps to be able to fulfill many requests without double-spending or data invalidation - SigmaUSD dApp can use proxy contracts to avoid bank double-spending and ERG/USD oracle data invalidation.

Eip-17 is proposed to define a solid structure for such contracts.

### DIFF
--- a/eip-0017.md
+++ b/eip-0017.md
@@ -8,21 +8,21 @@
 
 ## Motivation 
 
-Outsourcing transaction generation to an external service/dApp can be useful in various circumstances. For example, avoiding wallet limitations to generate any transaction on behalf of the user - [Ergo Assembler](https://github.com/anon-real/ergo-assembler) is designed for this purpose. Another example is to scale dApps to be able to fulfill many requests without double spending or data invalidation - SigmaUSD dApp can use proxy contracts to avoid bank double spending and ERG/USD oracle data invalidation within time.
+Outsourcing transaction generation to an external service/dApp can be useful in various circumstances. For example, avoiding wallet limitations to generate any transaction on behalf of the user - [Ergo Assembler](https://github.com/anon-real/ergo-assembler) is designed for this purpose. Another example is to scale dApps to be able to fulfill many requests without double-spending or data invalidation - SigmaUSD dApp can use proxy contracts to avoid bank double-spending and ERG/USD oracle data invalidation within time.
 
 
 ## Background
-The idea of proxy contracts came to life with the [Ergo Assembler](https://github.com/anon-real/ergo-assembler) which helped dApp developments like [Ergo Auction House](https://ergoauctions.org/), [ErgoUtils](https://ergoutils.org/) and [SigmaUSD web interface](https://sigmausd.io/#/) despite not having a wallet bridge like MetaMask (Ethereum wallet) in the ecosystem.
+The idea of proxy contracts came to life with the [Ergo Assembler](https://github.com/anon-real/ergo-assembler) which helped dApp developments like [Ergo Auction House](https://ergoauctions.org/), [ErgoUtils](https://ergoutils.org/), and [SigmaUSD web interface](https://sigmausd.io/#/) despite not having a wallet-bridge like MetaMask (Ethereum wallet) in the ecosystem.
 
 During this time, the structure of proxy contracts evolved as some malicious users tried to take advantage of some minor vulnerabilities, mostly in the [SigmaUSD web interface](https://sigmausd.io/#/).
 
 ## The structure
-At the beggining, the sole purpose of proxy contracts were to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infraustructure to work without malicious activities. 
+In the beginning, the sole purpose of proxy contracts was to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infrastructure to work without malicious activities. 
 
-For example, a malicious whale tried to take advantage of this simple structure by stealing UI fee from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities not the dApp infraustructure.
+For example, a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
 
 The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
 
 Generally, proxy contracts should be designed to:
 - prevent dApp developers or any other attacker from taking advantage of user's funds in any manner
-- preserve the integrity of the dApp by preventing attacks like ones explained in above examples.
+- preserve the integrity of the dApp by preventing attacks like the ones explained in the above examples.
--- a/eip-0017.md
+++ b/eip-0017.md
@@ -56,7 +56,7 @@ To achieve all of the above, the below contract structure is proposed as an exam
   sigmaProp(dAppWorksFine || returnFunds) // either dApp must work as it is supposed to or user's funds must be returned
 }
 ```
-The above contract is the proxy contract of minting operation in the [SigmaUSD dApp](https://sigmausd.io/#/). Anything started with `$` in the contract must be provided to it based on the user's request at the time of the compile.
+The above contract is the proxy contract of minting operation in the [SigmaUSD dApp](https://sigmausd.io/#/). Anything started with `$` in the contract must be provided to it based on the user's request at the time of compilation.
 
 The contract has three main parts:
 - Ensuring proper usage of user's funds, i.e., user will receive what he is paying for in proper amount without anyone being able to cheat.
--- a/eip-0017.md
+++ b/eip-0017.md
@@ -19,7 +19,7 @@ During this time, the structure of proxy contracts evolved as some malicious use
 ## The structure
 In the beginning, the sole purpose of proxy contracts was to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infrastructure to work without malicious activities. Some examples of dApp infrastructure violations are as follows:
 
-- For example, a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
+- A a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
 
 - The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
 
--- a/eip-0017.md
+++ b/eip-0017.md
@@ -1,0 +1,28 @@
+# Proxy contracts
+
+* Author: anon_real
+* Status: Proposed
+* Created: 05-May-2019
+* License: CC0
+* Forking: not needed 
+
+## Motivation 
+
+Outsourcing transaction generation to an external service/dApp can be useful in various circumstances. For example, avoiding wallet limitations to generate any transaction on behalf of the user - [Ergo Assembler](https://github.com/anon-real/ergo-assembler) is designed for this purpose. Another example is to scale dApps to be able to fulfill many requests without double spending or data invalidation - SigmaUSD dApp can use proxy contracts to avoid bank double spending and ERG/USD oracle data invalidation within time.
+
+
+## Background
+The idea of proxy contracts came to life with the [Ergo Assembler](https://github.com/anon-real/ergo-assembler) which helped dApp developments like [Ergo Auction House](https://ergoauctions.org/), [ErgoUtils](https://ergoutils.org/) and [SigmaUSD web interface](https://sigmausd.io/#/) despite not having a wallet bridge like MetaMask (Ethereum wallet) in the ecosystem.
+
+During this time, the structure of proxy contracts evolved as some malicious users tried to take advantage of some minor vulnerabilities, mostly in the [SigmaUSD web interface](https://sigmausd.io/#/).
+
+## The structure
+At the beggining, the sole purpose of proxy contracts were to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infraustructure to work without malicious activities. 
+
+For example, a malicious whale tried to take advantage of this simple structure by stealing UI fee from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities not the dApp infraustructure.
+
+The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
+
+Generally, proxy contracts should be designed to:
+- prevent dApp developers or any other attacker from taking advantage of user's funds in any manner
+- preserve the integrity of the dApp by preventing attacks like ones explained in above examples.
--- a/eip-0017.md
+++ b/eip-0017.md
@@ -2,19 +2,19 @@
 
 * Author: anon_real
 * Status: Proposed
-* Created: 05-May-2019
+* Created: 05-May-2021
 * License: CC0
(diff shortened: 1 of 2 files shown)


eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+- prevent dApp developers or any other attacker from taking advantage of user's funds in any manner
+- preserve the integrity of the dApp by preventing attacks like the ones explained in the above examples.
+
+To achieve all of the above, the below contract structure is proposed as an example:
+
+```scala
+{
+  // dApp-specific part ensuring that user will receive what he is paying for
+  val properFundUsage = {
+    val userOUt = OUTPUTS(1)
```
kushti: userOUt => userOut ?

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+    userOut.propositionBytes == fromBase64("$userAddress") && // user must be the recipient
+      myOut.tokens(0)._1 == fromBase64("$scTokenId") && // user must receive SigmaUSD
```
kushti: what is myOut ?

kushti (approved):
LGTM but please check the comments

greenhat (approved):

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+  val dAppWorksFine = properFundUsage && UIFeeOk && properBank
+
+  // in any case, whether assembler refuses to execute the request or the request fails for any reason, user must be able to get back his funds
+  val returnFunds = { 
+    val total = INPUTS.fold(0L, {(x:Long, b:Box) => x + b.value}) - $returnFee // only refund transactions's fee must be deducted from user's funds
+    OUTPUTS(0).value >= total && OUTPUTS(0).propositionBytes == fromBase64("$userAddress") // user must receive the appropriate amount
+    && (PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold) // either dApp-specific node can return user's funds or some time (block) has to be passed first. This is useful for many reasons.
+  }
+
+  sigmaProp(dAppWorksFine || returnFunds) // either dApp must work as it is supposed to or user's funds must be returned
```
nitram147, line 60: You are calling _**dAppWorksFine**_ in the OR manner.
dAppWorksFine is composed of _**properFundUsage && UIFeeOk && properBank**_
_**properFundUsage**_ contains (require thanks to the AND used) _**OUTPUTS.size == 2**_
_**UIFeeOk**_ contains (require thanks to the AND used) _**OUTPUTS.size == 4**_
Hence that means that only one or none member of tupple (_**properFundUsage**_, _**UIFeeOk**_) could be satisfied in some specific moment - because the _**OUTPUTS.size**_ has constant value during the contract execution.
This implies that the _**dAppWorksFine**_ could never be satisfied (AND used).

eip-0017.md
scalahub: I think its a typo. It should be userOut if I'm not mistaken

eip-0017.md
scalahub: yes this looks like a bug. I believe the `OUTPUTS.size == 2` should be part of `returnFunds` instead of `properFundUsage`

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+  // ensuring dApp integrity is preserved - any dApp specific condition to ensure designed procedures won't be violated
+  val UIFeeOk = OUTPUTS(2).propositionBytes == fromBase64("$implementor") && OUTPUTS.size == 4 // UI fee must go to UI devs not any random person who assembles the transaction
+  val properBank = OUTPUTS(0).tokens(2)._1 == fromBase64("$bankNFT") // the real bank box of the sigmaUSD protocol must be used so not any random person can behave as the bank box
+  val dAppWorksFine = properFundUsage && UIFeeOk && properBank
+
+  // in any case, whether assembler refuses to execute the request or the request fails for any reason, user must be able to get back his funds
+  val returnFunds = { 
+    val total = INPUTS.fold(0L, {(x:Long, b:Box) => x + b.value}) - $returnFee // only refund transactions's fee must be deducted from user's funds
+    OUTPUTS(0).value >= total && OUTPUTS(0).propositionBytes == fromBase64("$userAddress") // user must receive the appropriate amount
+    && (PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold) // either dApp-specific node can return user's funds or some time (block) has to be passed first. This is useful for many reasons.
```
scalahub: Minor comment. Should we allow attacker to initiate refund after threshold? If not then instead of `||` use `&&` for `HEIGHT` check. Also add  `PK(user)` so user can also get funds back in case assembler refuses to refund. So maybe the following
`&& ((PK("$assemblerNodeAddr") || PK("$useraddress")) && HEIGHT > $refundHeightThreshold)`; maybe threshold not needed here

eip-0017.md
anon-real: Typo, fixed.

eip-0017.md
anon-real: Right, last-minute changes broke the contract, fixed.

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+
+* Author: anon_real
+* Status: Proposed
+* Created: 05-May-2021
+* License: CC0
+* Forking: not needed 
+
+## Motivation 
+
+Outsourcing transaction generation to an external service/dApp can be useful or needed in various circumstances. For example, avoiding wallet limitations to generate any transaction on behalf of the user - [Ergo Assembler](https://github.com/anon-real/ergo-assembler) is designed for this purpose. Another example is to scale dApps to be able to fulfill many requests without double-spending or data invalidation - SigmaUSD dApp can use proxy contracts to avoid bank double-spending and ERG/USD oracle data invalidation.
```
scalahub: Better written as "Outsourcing transaction generation to an external service/dApp can be useful or **even** needed in various circumstances"

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+
+During this time, the structure of proxy contracts evolved as some malicious users tried to take advantage of some minor vulnerabilities, mostly in the [SigmaUSD dApp](https://sigmausd.io/#/).
+
+## The structure
+In the beginning, the sole purpose of proxy contracts was to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infrastructure to work without malicious activities. 
+
+For example, a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
```
scalahub: Not very clear what is "dApp infrastructure". Perhaps, its better to mention "UI provider"

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+
+The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
+
+Generally, proxy contracts should be designed to:
+- prevent dApp developers or any other attacker from taking advantage of user's funds in any manner
+- preserve the integrity of the dApp by preventing attacks like the ones explained in the above examples.
+
+To achieve all of the above, the below contract structure is proposed as an example:
+
```
scalahub, line 35: This is actually very specific to SigmaUSD. So maybe its better to call the EIP as SigmaUSD proxy contracts?

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+  val UIFeeOk = OUTPUTS(2).propositionBytes == fromBase64("$implementor") && OUTPUTS.size == 4 // UI fee must go to UI devs not any random person who assembles the transaction
+  val properBank = OUTPUTS(0).tokens(2)._1 == fromBase64("$bankNFT") // the real bank box of the sigmaUSD protocol must be used so not any random person can behave as the bank box
+  val dAppWorksFine = properFundUsage && UIFeeOk && properBank
+
+  // in any case, whether assembler refuses to execute the request or the request fails for any reason, user must be able to get back his funds
+  val returnFunds = { 
+    val total = INPUTS.fold(0L, {(x:Long, b:Box) => x + b.value}) - $returnFee // only refund transactions's fee must be deducted from user's funds
+    OUTPUTS(0).value >= total && OUTPUTS(0).propositionBytes == fromBase64("$userAddress") // user must receive the appropriate amount
+    && (PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold) && // either dApp-specific node can return user's funds or some time (block) has to be passed first. This is useful for many reasons.
+    OUTPUTS.size == 2 // only refund box and transaction fee box is needed
```
scalahub, line 57: What will this achieve: `OUTPUTS.size == 2`?
As long as user is getting total funds, he should not care what else is there

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+  }
+  
+  // ensuring dApp integrity is preserved - any dApp specific condition to ensure designed procedures won't be violated
+  val UIFeeOk = OUTPUTS(2).propositionBytes == fromBase64("$implementor") && OUTPUTS.size == 4 // UI fee must go to UI devs not any random person who assembles the transaction
+  val properBank = OUTPUTS(0).tokens(2)._1 == fromBase64("$bankNFT") // the real bank box of the sigmaUSD protocol must be used so not any random person can behave as the bank box
+  val dAppWorksFine = properFundUsage && UIFeeOk && properBank
+
+  // in any case, whether assembler refuses to execute the request or the request fails for any reason, user must be able to get back his funds
+  val returnFunds = { 
+    val total = INPUTS.fold(0L, {(x:Long, b:Box) => x + b.value}) - $returnFee // only refund transactions's fee must be deducted from user's funds
```
scalahub, line 54: Also consider the use-case of returning multiple users funds in a single transaction using the following technique:
`INPUTS(output.R4[Int].get).id == SELF.id` (ref: https://www.ergoforum.org/t/ergoscript-design-patterns/222/15?u=scalahub)

That way the assembler can benefit a little by paying fee from only one user and keeping the rest for himself :)

eip-0017.md
```diff
@@ -0,0 +1,70 @@
...
+
+The contract has three main parts:
+- Ensuring proper usage of user's funds, i.e., user will receive what he is paying for in proper amount without anyone being able to cheat.
+- Ensuring the integrity of the dApp procesures.
+- Ensuring that user will be refunded in any case of failures.
+
+
+Now we will go through some parts of the above contract which are both important and ambiguous at the first glance.
+
+- `(PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold)`: Either dApps's specific node can refund the user (at any time) or some time has to be passed for refunding without any secrets involved. This part prevents malicious users from sending refunds to users, preventing user's request (minting, in this case) from being executed. The same logic can be used to only allow a certain (or multiple) entities to execute requests. This enables dApps to control the order of execution which makes scaling dApps possible as it allows them to control doulbe-spending and other similar problems.
```
scalahub: typo in "double-spending"

eip-0017.md
anon-real: After the threshold, anyone can send the refund. `HEIGHT > $refundHeightThreshold)` is for when the assembler refuses to send the refund; this way, an attacker can not prevent the assembler from fulfilling the request, and at the same time the funds are safe.
To bind the contract to the user's address is a bit contradictory to what proxy contracts are.

eip-0017.md
anon-real: Fixed.

eip-0017.md
anon-real: The aim of the EIP is to propose a structure for proxy contracts and discuss some aspects/attacks for them. It does so with the SigmaUSD proxy contract because has evolved a lot through time with the whale's attacks. I think the discussed elements are useful for any proxy contracts (like for DEX).
If you think one proxy contract is not enough, I can add more.

eip-0017.md
anon-real: The whale executed an attack in which he sent the majority of the fee to himself. This caused transactions to be in the mempool for a long time. IMO, the more strict the contracts, the better.

eip-0017.md
anon-real: Good idea. But it also incentivizes the assembler to 1. do refund instead of executing the request. 2. wait for more user requests to fail and then send the refund to all of them which causes users to wait more.

eip-0017.md
anon-real: Added more examples. Should be clear now.

eip-0017.md
anon-real: Fixed.

eip-0017.md
```diff
@@ -17,11 +17,15 @@ The idea of proxy contracts came to life with the [Ergo Assembler](https://githu
...
+In the beginning, the sole purpose of proxy contracts was to protect users from losing their funds (not to be cheated) when they outsource their assets to engage with some dApp. While the initial structure succeeded to achieve this, it proved to be not sufficient for the whole dApp infrastructure to work without malicious activities. Some examples of dApp infrastructure violations are as follows:
 
-For example, a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
+- A a malicious whale tried to take advantage of this simple structure by stealing UI fees from SigmaUSD web interface developers for some period of time. This happened because the proxy contracts were simply only trying to protect users from malicious activities, not the dApp infrastructure.
```
nitram147: Typo, fix: "A a malicious" => "A malicious"

eip-0017.md
```diff
@@ -17,11 +17,15 @@ The idea of proxy contracts came to life with the [Ergo Assembler](https://githu
...
 
-The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
+- The same whale tried to prevent user's requests (minting/redeeming) from being executed by the assemble service by retuning their funds as soon as funds were broadcasted in the network. This happened also because of the same reasons.
```
nitram147: Typo, fix: "assemble service" => "assembly service"

eip-0017.md
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  sigmaProp(dAppWorksFine || returnFunds) // either dApp must work as it is supposed to or user's funds must be returned
+}
+```
+The above contract is the proxy contract of minting operation in the [SigmaUSD dApp](https://sigmausd.io/#/). Anything started with `$` in the contract must be provided to it based on the user's request at the time of compilation.
+
+The contract has three main parts:
+- Ensuring proper usage of user's funds, i.e., user will receive what he is paying for in proper amount without anyone being able to cheat.
+- Ensuring the integrity of the dApp procesures.
```
nitram147: Typo, fix: "dApp procesures" => "dApp procedures"

nitram147 (commented):
Found a few typos :)

eip-0017.md
anon-real: Fixed

eip-0017.md
anon-real: Fixed

eip-0017.md
anon-real: Fixed

eip-0017.md
```diff
@@ -0,0 +1,74 @@
...
+  val UIFeeOk = OUTPUTS(2).propositionBytes == fromBase64("$implementor") && OUTPUTS.size == 4 // UI fee must go to UI devs not any random person who assembles the transaction
+  val properBank = OUTPUTS(0).tokens(2)._1 == fromBase64("$bankNFT") // the real bank box of the sigmaUSD protocol must be used so not any random person can behave as the bank box
+  val dAppWorksFine = properFundUsage && UIFeeOk && properBank
+
+  // in any case, whether assembler refuses to execute the request or the request fails for any reason, user must be able to get back his funds
+  val returnFunds = { 
+    val total = INPUTS.fold(0L, {(x:Long, b:Box) => x + b.value}) - $returnFee // only refund transactions's fee must be deducted from user's funds
+    OUTPUTS(0).value >= total && OUTPUTS(0).propositionBytes == fromBase64("$userAddress") // user must receive the appropriate amount
+    && (PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold) && // either dApp-specific node can return user's funds or some time (block) has to be passed first. This is useful for many reasons.
```
nitram147, line 56: Wouldn't be better to allow also a user to issue refund transaction at any time (same way as the assembler service could).
So
`&& (PK("$assemblerNodeAddr") || HEIGHT > $refundHeightThreshold)`
would became
`&& ((PK("$assemblerNodeAddr") || PK("$userAddress")) || HEIGHT > $refundHeightThreshold)`
Is there any reason why we don't want the user to have right to do this before the **_$refundHeightThreshold_** threshold has elapsed? :)

anon-real: Explained here: https://github.com/ergoplatform/eips/pull/30#discussion_r640751675

nitram147: Sorry, I've missed it :)
So far everything LGTM :+1:

scalahub (approved):

PR #33, opened 2021-06-10
[WIP] EIP-0018: ErgoFund contracts

kushti:


eip-0018.md
```diff
@@ -0,0 +1,161 @@
...
+      "R7": "04a0be49",
+      "R8": "0580d0acf30e"
+    }
+  }
+]
+```
+
+## Pledge Contract
+
+Pledge contract is an example of self-sovereign DeFi principle. The contract is allowing box being protected by it to be spent if a spending transaction can collect at least *minToRaise* ERGs to *projectPubKey*, and inclusion height of the transaction is less than *deadline*. Otherwise, after *deadline* height is met, funds can be claimed by *projectPubKey*. Please note that both *backerPubKey* and
```
code-for-uss, line 130: I think funds can be claimed by *projectPubKey* Should be, funds can be claimed by *backerPubKey*

PR #35, opened 2021-07-13
Add N2T contracts

scalahub:
Contracts for LP and Sell Ergs.
Remaining contracts WIP

### DIFF
--- a/eip-0014.md
+++ b/eip-0014.md
@@ -672,17 +672,17 @@ PoolNFT             | Coll[Byte] | ID of the pool NFT (Used as a reference to a 
 
     val validTrade =
         OUTPUTS.exists { (box: Box) => // box containing the purchased tokens and balance of Ergs
-            val quoteAsset         = box.tokens(0)
-          
-            val quoteAssetID       = quoteAsset._1
-            val quoteAssetAmount   = quoteAsset._2
-          
-            val fairDexFee    = box.value >= SELF.value - quoteAssetAmount * DexFeePerTokenNum / DexFeePerTokenDenom - SellAmount
-
-            val relaxedOutput = quoteAssetAmount + 1 // handle rounding loss
-            val fairPrice     = poolReservesY.toBigInt * SellAmount * FeeNum <= relaxedOutput * (poolReservesX.toBigInt * FeeDenom + SellAmount * FeeNum)
-
-            val uniqueOutput = INPUTS(box.R4[Int].get).id == SELF.id // See https://www.ergoforum.org/t/ergoscript-design-patterns/222/15?u=scalahub            
+          val quoteAsset         = box.tokens(0)
+        
+          val quoteAssetID       = quoteAsset._1
+          val quoteAssetAmount   = quoteAsset._2
+        
+          val fairDexFee    = box.value >= SELF.value - quoteAssetAmount * DexFeePerTokenNum / DexFeePerTokenDenom - SellAmount
+  
+          val relaxedOutput = quoteAssetAmount + 1 // handle rounding loss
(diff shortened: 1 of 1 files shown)


eip-0014.md
```diff
@@ -510,3 +510,239 @@ PoolNFT        | Coll[Byte] | Pool NFT ID
     sigmaProp(Pk || (validPoolIn && validReturnOut))
 }
 ```
+
+### Ergo AMM DEX Contracts [Ergo to Token]
+
+The Ergo-to-token exchange is essentially the native-to-token (N2T) contract of UniSwap 1.0.
```
oskin1: Comparison with UniSwap 1.0 isn't 100% accurate. UniSwap 1.0 has different initialization formulas, doesn't have LP burning, variable fee.

eip-0014.md
```diff
@@ -510,3 +510,239 @@ PoolNFT        | Coll[Byte] | Pool NFT ID
...
+There are two approaches to create a N2T exchange:
+1. Use a T2T (token-to-token) exchange, where one of the tokens maps to Ergs and have a separate dApp that exchanges Ergs to tokens at 1:1 rate.
+2. Implement N2T directly in the exchange contract. Here we use this approach.
+
+The N2T AMM DEX relies on two types of contracts as before:
+
+- Pool contracts
+- Swap contracts
+
+#### Pool contracts
```
oskin1, line 521: Why repeating this part? I believe general protocol description can be unified. N2T and T2T are just implementations.

eip-0014.md
```diff
@@ -510,3 +510,239 @@ PoolNFT        | Coll[Byte] | Pool NFT ID
...
+tokens[2] | Asset Y reserves
+R4[Long]  | Fee multiplier numerator (e.g. 0.003% fee -> 997 fee_num). This represents the *non-fee* part of the sold asset
+
+#### Pool contract
+
+```scala
+{
+    val InitiallyLockedLP = 0x7fffffffffffffffL
+    val FeeDenom = 1000
+    val minStorageRent = 10000000L  // this many number of nanoErgs are going to be permanently locked
```
oskin1, line 553: Why permanently locked? Why LP can't redeem all the liquidity and destroy the pool? Otherwise unless we consider this locked reserves in deposit / redeem formulas there will be some portion of LP tokens that can never be returned.

eip-0014.md
```diff
@@ -510,3 +510,239 @@ PoolNFT        | Coll[Byte] | Pool NFT ID
...
+DexFeePerTokenNum   | Long       | Numerator of the DEX fee in nanoERGs per one unit of quote asset
+DexFeePerTokenDenom | Long       | Denominator of the DEX fee in nanoERGs per one unit of quote asset
+PoolNFT             | Coll[Byte] | ID of the pool NFT (Used as a reference to a concrete unique pool)
+
+```scala
+{   // contract to sell tokens and buy Ergs
+    val FeeDenom = 1000
+    
+    val sellToken = SELF.tokens(0)
+    val sellTokenId = sellToken._1
```
oskin1: Lets preserve terminology from T2T contracts.

oskin1 (changes requested):
Also, please add yourself to the authors list.

eip-0014.md
scalahub: Redeem complete liquidity and destroying the pool can never happen even in the T2T contracts. The moment you have successor condition, the contract will fail if there is no successor, and so we will just let that small amount of LP stuck till miner can claim it. If we can modify the normal T2T contracts to accommodate destruction, we can port same logic here. We can think of adding more checks to allow complete LP redemption but that will make it more complex, so I suggest we leave it.

oskin1 (approved):

PR #36, opened 2021-08-18
EIP-0019: Cold Wallet

aslesarenko:
This standard, when supported by mobile wallets, can be used to create transactions in the Hot wallet (connected to network), pass them via QR codes to Cold Wallet (disconnected from network) for signing and then receiving the signed transaction back.
Benefits:
- Hot Wallet don't need private keys
- Cold Wallet don't need network connections and don't need to implement script reduction (only signature implementation is required)

### DIFF
--- a/eip-0019.md
+++ b/eip-0019.md
@@ -182,7 +182,7 @@ ColdSigningRequest data format and Cold Wallet replies back with ColdSigningResp
 Json format, holding
 * reducedTx (mandatory): Base64-encoded `ReducedTransaction` as defined above
 * sender (optional): P2PK address sending the transaction, can be used by cold wallets to determine which secret to use for signing the transaction
-* inputs (optional): List of base64-encoded serialized input boxes, can be used by cold wallets to show the user which boxes are burnt
+* inputs (mandatory): List of base64-encoded serialized input boxes, can be used by cold wallets to show the user which boxes are burnt
 
 ### ColdSigningResponse:
 Json format, holding
--- a/eip-0019.md
+++ b/eip-0019.md
@@ -198,12 +198,11 @@ As QR codes are length-limited and it could be needed to transfer the data in ch
 Examples:
 The QR code for a complete ColdSigningRequest contains the following data:
 
-     {"reducedTx":"....","sender":"9...","inputs":["...."]}
      {"CSR":"{\"reducedTx\":\"....\",\"sender\":\"9...\",\"inputs\":[\"...\"]}"}
 
 The QR code for the first chunk of a ColdSigningResponse with 3 pages total looks like the following:
 
-     {"CSTX":"{\"signedTx\":\"...\",\"s","n":3,"p":1}
+     {"CSTX":"{\"signedTx\":\"... (actual data, no valid JSON on its own)","n":3,"p":1}
 
 
 In addition to using the above formats the following requirements are imposed on
--- a/eip-0019.md
+++ b/eip-0019.md
@@ -135,7 +135,7 @@ method.
 This is because those context data is not required to generate proof (aka signature)
 once ErgoTree is reduced to ReductionResult containing sigma proposition.
 
-Having the `ReducedTransaction` data structure the the full transaction signing
+Having the `ReducedTransaction` data structure the full transaction signing
 consists of three steps
 
 1) Create unsigned transaction and then reduce it in the current
--- a/eip-0019.md
+++ b/eip-0019.md
@@ -101,17 +101,20 @@ serialization format called `ReducedTransaction`.
 ReducedTransaction:
   - unsignedTx: UnsignedTransaction
   - reducedInputs: Seq[ReducedInputData]
+  - txCost: Int
 
 UnsignedTransaction:
   - inputs: Seq[UnsignedInput],
   - dataInputs: Seq[DataInput],
-  - outputCandidates: Seq[ReductionResult]
+  - outputCandidates: Seq[ErgoBoxCandidate]
 
 UnsignedInput:
   - boxId: BoxId
   - extension: ContextExtension
 ```
 
+ReducedInputData:
+  - reductionResult: ReductionResult 
 Thus, the `ReducedTransaction` instance contains unsigned transaction augmented with
 one `ReductionResult` for each `UnsignedInput`. 
 
--- a/eip-0019.md
+++ b/eip-0019.md
@@ -9,6 +9,11 @@
 ## Contents
 - [Description](#description)
 - [Background And Motivation](#background-and-motivation)
+- [Reference Implementation](#reference-implementation)
+  - [Implementation in Mobile Wallet](#implementation-in-mobile-wallet)
+  - [Implementation in dApp](#implementation-in-dapp)
+- [Benefits for dApps](#benefits-for-dapps)
+- [Benefits for Mobile Wallets](#benefits-for-mobile-wallets)
 
 ## Description 
 This EIP defines a standard for cross-platform interaction between an online dApp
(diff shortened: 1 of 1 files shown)


eip-0019.md
```diff
@@ -0,0 +1,221 @@
...
+reduction and which are part of the
+[Context](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e5127f6743db824f7280881cd5c4ecd336075e2f/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala#L51)
+data structure required by the
+[prove](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/f24833d8d4572d77e4a93e5b69360335cb2d7dc1/sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala#L104)
+method.
+
+This is because those context data is not required to generate proof (aka signature)
+once ErgoTree is reduced to ReductionResult containing sigma proposition.
+
+Having the `ReducedTransaction` data structure the the full transaction signing
```
greenhat:
```suggestion
Having the `ReducedTransaction` data structure the full transaction signing
```

MrStahlfelge:
Reference implementation is done, we can convert this to ready for review when you think it is done @aslesarenko

eip-0019.md
```diff
@@ -0,0 +1,238 @@
...
+
+```
+ReducedTransaction:
+  - unsignedTx: UnsignedTransaction
+  - reducedInputs: Seq[ReducedInputData]
+
+UnsignedTransaction:
+  - inputs: Seq[UnsignedInput],
+  - dataInputs: Seq[DataInput],
+  - outputCandidates: Seq[ReductionResult]
```
greenhat:
```suggestion
  - outputCandidates: Seq[ErgoBoxCandidate]
```

eip-0019.md
```diff
@@ -0,0 +1,238 @@
...
+
+UnsignedInput:
+  - boxId: BoxId
+  - extension: ContextExtension
+```
+
```
greenhat, line 115:
```suggestion

ReducedInputData:
  - reductionResult: ReductionResult 
```
based on appkit implementation

eip-0019.md
```diff
@@ -175,16 +175,33 @@ For better security and usability additional data can be transfered between Hot
...
- - description: String
-
-ColdSigningResponse:
- - messageType: Byte = 2
- - transaction: SignedTransaction
-```
+### ColdSigningRequest
+Json format, holding
+* reducedTx (mandatory): Base64-encoded `ReducedTransaction` as defined above
+* sender (optional): P2PK address sending the transaction, can be used by cold wallets to determine which secret to use for signing the transaction
```
aslesarenko, line 184: In the case of EIP3 addresses, many secret keys may need to be added to the prover in order to sign the transaction.
So, it is better to make this field a list (json array).

eip-0019.md
```diff
@@ -175,16 +175,33 @@ For better security and usability additional data can be transfered between Hot
...
+
+* Constant CSR for ColdSigningRequest or CSTX for ColdSigningResponse
+* (if pages num > 1) const / and 1-based page index of current chunk
+* (if pages num > 1) const / and pages num
+* const -
+
+Examples:
+The QR code for a complete ColdSigningRequest contains the following data:
+
+     CSR-{"reducedTx":"....","sender":"9...","inputs":["...."]}
```
aslesarenko: I think it is better to make the QR code contents a valid Json document.
For example
```
{"CSR": {"reducedTx":"....", "sender":"9...", "inputs":["...."]}}
{"CSTX": {p: 1, n:3, "signedTx":"..."}}
```
The will be easier to both parse and also understand.

eip-0019.md
MrStahlfelge: Not sure if this is needed. My plan was to just send the first address, and the cold wallets should add all derived addresses that are configured. This keeps the payload small.

eip-0019.md
MrStahlfelge: I agree, it is easier to implement that way. I aimed to keep the payload as small as possible, but when using just p and n this is small as well.

eip-0019.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Simplified Signing using ReducedTransaction
+
+To solve the problem of _cold signing_ we need a new data structure and
+serialization format called `ReducedTransaction`.
+
+```
+ReducedTransaction:
+  - unsignedTx: UnsignedTransaction
+  - reducedInputs: Seq[ReducedInputData]
```
greenhat, line 103: Following our discussion on Discord and keeping in mind that appkit already has it implemented:
```suggestion
  - reducedInputs: Seq[ReducedInputData]
  - txCost: Int
```

eip-0019.md
aslesarenko: There is one subtlety here. ReducedInputData indeed has ContextExtension, however it is not serialized because it is already serialized as part of Input.

anon-real (commented):
LGTM as well. However, there are some ambiguities for me.
You have mentioned the Ergo Wallet Android as the reference implementation of cold and hot wallets. I understand that cold/hot wallets do not necessarily need to be separated but this makes the EIP description a bit weird. In this case, dapp that is requesting a tx to be signed is actually a hot wallet based on the EIP. 

In that case, I think it is better not to constraint the way hot and cold wallets interact since QR codes are limited. If they are on the same device, they may interact in some other fashion like between-app-interactions that Android allows.

aslesarenko:
@anon-real could you please point to the line in the text where this ambiguity starts.

MrStahlfelge:
@anon-real maybe you confused this with EIP-0020 ErgoPay? This one is not for dapps but only for interaction between cold wallets and hot wallets.

anon-real:
@aslesarenko @MrStahlfelge what I mean is that the combination of this EIP and EIP-0020 does not make sense to me.

[Here](https://github.com/ergoplatform/eips/pull/36/files#diff-23bce28f1d00b2cdbcb0eaa6912803d5fec6dc250aac695ec35a1038eb3221d1R144) you are describing how the procedure works in general. The hot wallet serializes the reduced transaction and ... However, in EIP-0020, the entity that does this procedure is the dapp. And what adds more to the confusion is that the hot and cold implementations reference are in the same wallet. There is not a clear distinction between the dapp, hot wallet, and cold wallet and the interaction between them in these two EIPs.

aslesarenko:
@anon-real for this EIP we already have a reference implementation. Yes both hot and cold wallets are in the same github repo, but still there is an implementation which implement both Hot and Cold wallet roles. This are roles, not apps. 

The goal of this EIP is to fix the interaction protocol so that, for example, alternative wallets could use Android Cold Wallet reference implementation.

The EIP-20 is not implemented yet and no doubt there may be (and in fact should be) some changes. The first pair of implementations (dApp + mobile wallet) may have clear reasons and motivated suggestions for the change in EIP-20.

From a high level. 
- EIP19 is about signing a ReducedTransaction.
- EIP20 is about how to get the ReducedTransaction, and submit SignedTransaction and not about the details of signing it.

eip-0019.md
MrStahlfelge: It's now wrapped into JSON instead of an own prefix. Please note that the inner JSON has to be a String and not a JSON object due to the fact that for chunks it isn't a valid JSON object.

MrStahlfelge:
Changed the QR code paging to a JSON format and made the ColdSigningRequest.inputs mandatory (as discussed [here](https://github.com/ergoplatform/ergo-wallet-app/pull/55#discussion_r753338055))

greenhat (approved):

MrStahlfelge (approved):
Concept has proven to work in practice

PR #37, opened 2021-08-25
EIP-0020: Ergo Pay

aslesarenko:
This standard defines how online dApps can interact with mobile wallets to sign and send transactions. Based on `ReducedTransaction` described in details in #36 .

### DIFF
--- a/eip-0020.md
+++ b/eip-0020.md
@@ -22,10 +22,10 @@ and a mobile wallet for creating, signing and sending Ergo transactions.
 
 ## Background And Motivation
 
-Mobile wallets, (like [Ergo Andorid Wallet]()) typically
+Cryptocurrency wallets, (like [Ergo Android Wallet](https://github.com/ergoplatform/ergo-wallet-android)) typically
 support scanning QR codes of [Payment
-Request](https://explorer.ergoplatform.com/payment-request?address=9er9hxmVcL8S4bNypCyJHpPEEkAfEd9CLq5gNrHN6s2pYomp55N&amount=0&description=).
-After scanning a payment request the wallet can build a new transaction, sign it
+Request](https://explorer.ergoplatform.com/payment-request?address=9er9hxmVcL8S4bNypCyJHpPEEkAfEd9CLq5gNrHN6s2pYomp55N&amount=0&description=) (on mobile devices) or intercepting a link with a special [URI scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki).
+After parsing a payment request the wallet can build a new transaction, sign it
 using a secret key stored locally on the device and then send to the blockchain.
 
 However, in general, this only can be done for _simple_ transactions like
@@ -36,7 +36,7 @@ The reason for this is simple. In Ergo's eUTXO model a box can be protected by
 an arbitrarily complex contract (aka spending condition) and any spending
 transaction should satisfy that condition by adding required context variables,
 creating expected number of outputs with specific registers depending on the
-contract. There is no way for a universal mobile wallet application to know all
+contract. There is no way for a universal wallet application to know all
 the specific details of all the possible contracts.
 
 On the contrary, every Ergo dApp (think of website) is usually built on top of
@@ -47,11 +47,13 @@ can spend its own contracts without above mentioned problems. However, the dApp
 cannot sign those transactions, because signing requires knowledge of private keys
 which are not stored on the dApp, but instead stored on the mobile wallet.
 
-Thus, interaction between a dApp and a mobile wallet is required such that:
-1) the dApp builds a transaction and makes it available for the mobile wallet
-2) mobile wallet signs the transaction and submits it to the blockchain or
+Thus, interaction between a dApp and a wallet is required such that:
+1) the dApp builds a transaction and makes it available for the wallet application
+2) wallet app shows a confirmation screen to the user, displaying the inboxes and 
+outboxes it is going to sign
+3) when user confirms, wallet app signs the transaction and submits it to the blockchain or
 returns back to the dApp
-3) dApp monitors the transaction on the blockchain and upon confirmations
+4) dApp monitors the transaction on the blockchain and upon confirmations
 proceeds with its business logic.
 
 One possible implementation of such an interaction scheme is described in
@@ -60,7 +62,7 @@ suitable for in-browser wallets, which can interact with dApps via javascript
 code injected into a web context.
 
 Luckily, the design of Ergo contracts allows for a simple and universal
-implementation suitable for mobile wallets. This EIP describes the interaction
+implementation suitable for mobile or desktop wallets. This EIP describes the interaction
 protocol and the [Reference Implementation](#reference-implementation).
 
 In this 3 step process the main question is: 
@@ -69,7 +71,7 @@ application so that it can be signed there?".
 
 ## ErgoPay Interaction Protocol
 
-An ErgoPay interaction between Hot Wallet, and dApp is driven by dApp's
+An ErgoPay interaction between Wallet, and dApp is driven by dApp's
 user and proceeds as the following:
 
 1) The user enters the necessary information in the dApp's UI and proceeds to a
@@ -77,7 +79,7 @@ payment screen.
 
 2) The payment screen shows transaction details and a QR code.
 
-3) User scans the QR code using a Hot Wallet application (see
+3) User scans the QR code using a Wallet application (see
 [EIP-0019](eip-0019.md))
 
 4) The Hot Wallet application parses the QR code data and obtains either
@@ -115,7 +117,7 @@ monitoring that transaction.
 
 ## Data Formats
 
-The data formats of this EIP are based on a new data structure and
+The data formats of this EIP are based on a new binary data structure and
 serialization format called `ReducedTransaction` which is described in
 [EIP-0019](eip-0019.md).
 
@@ -124,32 +126,34 @@ formats defined in EIP-0019.
 
 Depending on the bytes size of `ReducedTransaction` it may not fit into 4K limit
 of QR codes. In this case or for whatever reason the dApp may encode the
-following data in the QR code on the payment screen.
+following data in the QR code on the payment screen or link to it:
 
-```
-ErgoPaySigningRequestUrl
-  - messageType: Byte = 2  // request url
-  - url: String
-```
+`ergopayurl://<URL without https>`, for example
+`ergopayurl://sigmausd.io/signingRequest/2001-16b8-66c4-b800-6e52-8ce4`
 
-When the Hot Wallet application parses the QR code bytes it can use first byte
-(the `messageType` field) to learn the message type.
-When `ErgoPaySigningRequestUrl` is parsed, the Hot Wallet application should
-request dApp and obtain the following data.
+When the Wallet application parses the QR code bytes it can use prefix URI scheme
+to learn the message type.
+
+The wallet application should request URL and obtain the following data (json format)
 
 ```
 ErgoPaySigningRequest:
-  - messageType: Byte = 1  // request itself
   - transaction: ReducedTransaction
-  - description: String
-  - replyToUrl: String
+  - description: String (optional)
+  - replyToUrl: String (optional)
 ```
+
+When dapp does not make use of an extra request, the `ErgoPaySigningRequest` could 
+also be encoded in the QR code or link:
+
+`ergopay:<ReducedTransaction, base 64 url safe encoded>`
+
 Once the transaction details obtained, they CAN BE OPTIONALLY shown on the payment
 screen of the Hot Wallet application.
 
-After signing is performed and the `SignedTransaction` data is obtained, the Hot
-Wallet need to send the following data to the dApp using `replyToUrl` from the
-signing request.
+After signing is performed and the `SignedTransaction` data is obtained, the 
+wallet app need to POST the following data to the dApp using `replyToUrl` from the
+signing request (json format).
 
 ```
 ErgoPayTransactionId:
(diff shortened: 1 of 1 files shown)


eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+- [ErgoPay Interaction Protocol](#ergopay-interaction-protocol)
+- [Data Formats](#data-formats)
+- [Implementation in Mobile Wallet](#implementation-in-mobile-wallet)
+- [Implementation in dApp](#implementation-in-dapp)
+- [Benefits for dApps](#benefits-for-dapps)
+- [Benefits for Mobile Wallets](#benefits-for-mobile-wallets)
+
+## Description 
+This EIP defines a standard for cross-platform interaction between an online dApp
+and a mobile wallet for creating, signing and sending Ergo transactions.
```
MrStahlfelge: We can use this for desktop wallets as well, by using a link like ergopay://signingrequest?data=..., so the EIP could be generalized to mobile and desktop wallets.

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+specific contracts. Business logic of the dApp includes spending of the boxes
+protected by those application-specific contracts.
+In other words, any dApp is in full control of creating transactions and thus
+can spend its own contracts without above mentioned problems. However, the dApp
+cannot sign those transactions, because signing requires knowledge of private keys
+which are not stored on the dApp, but instead stored on the mobile wallet.
+
+Thus, interaction between a dApp and a mobile wallet is required such that:
+1) the dApp builds a transaction and makes it available for the mobile wallet
+2) mobile wallet signs the transaction and submits it to the blockchain or
```
MrStahlfelge: We should mention a user confirmation screen here (it is mentioned later)

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+
+An ErgoPay interaction between Hot Wallet, and dApp is driven by dApp's
+user and proceeds as the following:
+
+1) The user enters the necessary information in the dApp's UI and proceeds to a
+payment screen.
+
+2) The payment screen shows transaction details and a QR code.
+
+3) User scans the QR code using a How Wallet application (see
```
MrStahlfelge: How -> Hot

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+Hot Wallet containing the same transaction details as the dApp screen.
+
+6) The user compares the dApp's screen, the Wallet's screen and the
+transaction details and confirms the payment by using a "Sign" button. If the Hot
+Wallet also supports [EIP-0019](eip-0019.md) and private keys are not available,
+then the sign button behaves like a "Cold Sign" button according to EIP-0019. 
+
+7) The wallet application signs the transaction either using local private keys
+or using Cold Wallet and EIP-0019 protocol. The result of signing is
+`SigningTransaction` data.
```
MrStahlfelge: SignedTransaction?

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+the dApp
+- the Hot Wallet MAY implement "one-at-a-time" signing. When the signing
+is interrupted for some reason it can be started from the step 3) (i.e. scanning
+the QR)
+- when dApp receives the transaction id, but the `SignedTransaction` is not
+accepted to the blockchain then the dApp SHOULD detect this situation and stop
+monitoring that transaction.
+
+## Data Formats
+
```
MrStahlfelge, line 125: We should define a link format for desktops and for interception on a mobile device via tapping a link as well (like mentioned above ergopay://signingrequest?data=...)

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+request dApp and obtain the following data.
+
+```
+ErgoPaySigningRequest:
+  - messageType: Byte = 1  // request itself
+  - transaction: ReducedTransaction
+  - description: String
+```
+Once the transaction details obtained, they should be shown on the payment
+screen of the Hot Wallet application.
```
MrStahlfelge: This step should be optional.
It is not clear from where the information about the API endpoint comes from. It could be an optional field in `ErgoPaySigningRequestUrl`, `ErgoPaySigningRequest` or both.

eip-0020.md
```diff
@@ -0,0 +1,196 @@
...
+Pay button, without any interim screens or pop-ups except to prompt for
+necessary product details, such as size or quantity.
+- Ergo Pay is simple and universal. It supports all smart contracts and offers
+the flexibility to implement simple to complex dApps.
+
+## Benefits for Mobile Wallets
+
+Any mobile wallet team should consider supporting Ergo Pay in their wallet along
+with basic wallet features. Users can participate in Ergo dApps and mobile
+wallet team can receive service fees from those transactions.
```
MrStahlfelge, line 245: How can the wallet team receive service fees? The only way I can imagine would be by changing the transaction and add an additional outbox - this is probably not intended.

eip-0020.md
aslesarenko: I'm not sure I get the idea, please clarify.

eip-0020.md
aslesarenko: fixed

eip-0020.md
aslesarenko: fixed

eip-0020.md
aslesarenko: My original idea is that the data is not passed via link. The link is used to request the actual data from the server.

eip-0020.md
aslesarenko: fixed

eip-0020.md
aslesarenko: yes, in this way, it may be acceptable as long as the user aware of the additional fees.
We can describe this in more details as a second revision of the standard if necessary.

eip-0020.md
MrStahlfelge: When we restrict this to transmit the information via QR code, this can be only used when using both desktop and mobile device together. When used on a single mobile device or on a desktop, we need another way to start the process. This can be done with said link ergopay://signingrequest?data= which defines a new URI scheme to hand the data over to an application. Handlers can be registered for this [on Desktop](https://medium.com/swlh/custom-protocol-handling-how-to-8ac41ff651eb) and [on mobile](https://stackoverflow.com/a/11421251/7487013)

eip-0020.md
MrStahlfelge: Yeah, that is fine and we should leave this. My suggestion was to use a new URI scheme to start the wallet application on desktop or on mobile when QR code is not possible to use.

eip-0020.md
aslesarenko: @MrStahlfelge sounds good, please add the necessary description in the text.

eip-0020.md
aslesarenko: @MrStahlfelge please  add in the text.

eip-0020.md
```diff
@@ -0,0 +1,202 @@
...
+
+## ErgoPay Interaction Protocol
+
+An ErgoPay interaction between Wallet, and dApp is driven by dApp's
+user and proceeds as the following:
+
+1) The user enters the necessary information in the dApp's UI and proceeds to a
+payment screen.
+
+2) The payment screen shows transaction details and a QR code.
```
kushti: It is needed to show bytes also, then support can be added to desktop wallets as well

eip-0020.md
```diff
@@ -0,0 +1,202 @@
...
+proceeds with its business logic.
+
+One possible implementation of such an interaction scheme is described in
+[EIP-0012](https://github.com/ergoplatform/eips/pull/23). It, however, only
+suitable for in-browser wallets, which can interact with dApps via javascript
+code injected into a web context.
+
+Luckily, the design of Ergo contracts allows for a simple and universal
+implementation suitable for mobile or desktop wallets. This EIP describes the interaction
+protocol and the [Reference Implementation](#reference-implementation).
```
kushti: The link seems to be broken

eip-0020.md
MrStahlfelge: Either that, another way to go is to use an URI scheme `ergopay:<ReducedTransaction, base 64 url safe encoded>` mentioned earlier in the document. This way, browser can hand over the data to desktop wallets automatically, and it works on mobile browser to mobile wallet as well.

eip-0020.md
```diff
@@ -0,0 +1,202 @@
...
+
+8) The Hot Wallet obtains the transaction id and sends it to the dApp using
+`ErgoPayTransactionSent` API post message, if successful the wallet then submits
+`SignedTransaction` to the blockchain.
+
+9) dApp monitors the transaction by id and proceeds with its business logic upon
+receiving enough confirmations. This concludes the ErgoPay protocol.
+
+Additional requirements: 
+- many instances of ErgoPay protocol MUST be able to run simultaneously withing
```
anon-real: Typo: "withing"

eip-0020.md
```diff
@@ -0,0 +1,202 @@
...
+signing request (json format).
+
+```
+ErgoPayTransactionId:
+  - txId: String
+```
+
+After `ErgoPayTransactionId` is sent, only then the dApp can submit the
+SignedTransaction to the blockchain. The order is important, otherwise, if
+transaction is included into blockchain and dApp notification fails the dApp
```
anon-real: I don't understand the argument here.
First of all, not all dapps can support `replyToUrl` because they may simply be serverless (like the current AH, ErgoUtils and SigmaUSD) so they can not provide an API for that.
I think a more general solution is needed here. In the simplest case, the dapp can keep monitoring for the outputs that it expects to see if they are in the mempool.

anon-real: Also, if I remember correctly, an unsigned tx ID can be obtained. So I think the dapp can know the tx ID even before it being signed. @kushti can clarify this.

eip-0020.md
```diff
@@ -0,0 +1,202 @@
...
+mobile wallet guarantees they never leave the mobile device
+- ErgoPay EIP is compatible with Cold Wallet EIP, thus users can use Cold Wallet
+devices to sign transaction within ErgoPay signing process.
+- adding Ergo Pay to product detail pages, the cart, checkout page, in payment
+settings, or anywhere else a user can choose ErgoPay as the payment method or
+initiate a purchase.
+- The payment screen can be presented immediately after the user taps the Ergo
+Pay button, without any interim screens or pop-ups except to prompt for
+necessary product details, such as size or quantity.
+- Ergo Pay is simple and universal. It supports all smart contracts and offers
```
anon-real: Technically, I don't think it supports "all smart contracts" but most of them. For example, it can not support multi-sig contracts.

anon-real: Ok, I think clarification is needed here. Do you mean it supports any smart contract in the outputs or inputs?
The way I understand `ReducedTransaction` is that it reduces the script to a simple operation like `proveDlog` that any wallet supports, so I think you mean inputs here.

anon-real (changes requested):

eip-0020.md
aslesarenko: fixed

eip-0020.md
aslesarenko: @anon-real yes, indeed unsignedTx.id == signedTx.id, so this requirement can be removed.
I clarified this point in the text.

eip-0020.md
aslesarenko: ReducedTransaction contains inputs reduced to SigmaBoolean, i.e. any sigma protocol proposition including multi-signatures. So, the claim is correct, it is indeed universal solution.

anon-real (approved):

eip-0020.md
anon-real: Wow, this is exciting then.

eip-0020.md
MrStahlfelge: Removed

eip-0020.md
```diff
@@ -0,0 +1,246 @@
...
+link.
+
+3) User clicks the link to hand the information over to a wallet application on the 
+same device, or scans the QR code using a wallet application.
+
+4) The Hot Wallet application parses the QR code data and obtains either
+`ErgoPaySigningRequest` or `ErgoPaySigningRequestUrl` data (see [Data Formats](#data-formats) section). In the later case the Hot
+Wallet uses the scanned url to download `ErgoPaySigningRequest` from the dApp.
+
+5) When `ErgoPaySigningRequest` is obtained, it is shown as a payment screen on the
```
JimmyHoffa: I think there's a 5a and 5b here based on `option 1` and `option 2` below - that is to say for the serverless `option 2` below it would be good to clarify here you don't need to create the entire `ErgoPaySigningRequest`, rather just the url safe ReducedTx

eip-0020.md
```diff
@@ -0,0 +1,246 @@
...
+In case no **transaction** was provided, the wallet app displays the **message** that should inform the
+user about further steps needed.
+
+
+### Option 2: dApp providing URI scheme containing reduced transaction
+
+When dapp does not need to use an extra request, the `ReducedTransaction` could 
+also be encoded in the QR code or link:
+
+`ergopay:<ReducedTransaction, base 64 url safe encoded>`
```
JimmyHoffa, line 200: maybe a note or a link to other docs that "url safe" denotes a replacement of `'+' = '-'` and `'/' = '_'`

eip-0020.md
MrStahlfelge: Clarified

eip-0020.md
MrStahlfelge: It's a standard encoding format, implemented in most programming languages. So I think we don't need a further definition here.

MrStahlfelge (approved):
Implemented in wallet app 1.6 and working for showcase app.
First dApp and other wallet implementations using it.

JimmyHoffa (approved):

ThierryM1212 (approved):
Seems great, implemented signing of b64 reduced transaction in SAFEW wallet. It seems to give a good user experience, tested with ergodex and tokenminter. A alpha will be added soon with the feature.

oskin1 (approved):

anon-real (approved):

anon-real (approved):

PR #38, opened 2021-08-31
EIP-0021 Genuine tokens verification

MrStahlfelge:
This EIP proposes a simple token verification for end-user applications

### DIFF
--- a/eip-0021.md
+++ b/eip-0021.md
@@ -49,10 +49,10 @@ Proposed warning sign: [Material icons dangerous](https://fonts.google.com/icons
 
 ### Genuine tokens
 
-| Verbose name       | Token id                                        | Unique name |
-| ------------------ |:-----------------------------------------------:| -----------:|
-| SigUSD             |                                                 | yes         |
-| SigRSV             |                                                 | yes         |
+| Verbose name       | Token id                                                         | Unique name | Issuer
+| ------------------ |:----------------------------------------------------------------:| -----------:| ------
+| SigUSD             | 03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04 | yes         | sigmausd.io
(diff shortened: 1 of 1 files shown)


eip-0021.md
```diff
@@ -0,0 +1,60 @@
...
+Proposed warning sign: [Material icons dangerous](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Adangerous%3A)
+
+## Token lists
+
+### Genuine tokens
+
+| Verbose name       | Token id                                        | Unique name |
+| ------------------ |:-----------------------------------------------:| -----------:|
+| SigUSD             |                                                 | yes         |
+| SigRSV             |                                                 | yes         |
```
aslesarenko: Not sure the `Unique name` column is necessary.
Also please put the actual ids for the tokens in the table.

aslesarenko (changes requested):

eip-0021.md
MrStahlfelge: Do you think unique name is not necessary because the name for tokens on the list should always be unique, or because you think warning about an impersonated token name is not necessary?

MrStahlfelge: Added the IDs

Luivatra:
The token space on ergo is exploding and with the bridges bringing in wrapped assets and dapps using their own it will be difficult for the average user to discern fake tokens from the real ones. A community curated list available through an api would help out a lot for dapps and wallets to give their users a good experience. Would this be seperate or part of fe. the explorer?

MrStahlfelge:
This list could be hardcoded into apps, but I think it would be great to have this on the explorer API. This way, changes on the list would affect all apps using the API directly.

greenhat (approved):

kushti (approved):

aslesarenko (approved):

PR #39, opened 2021-09-01
EIP-0022 Auction Contract

anon-real:
This EIP proposes the auction contract.

### DIFF
--- a/eip-0022.md
+++ b/eip-0022.md
@@ -1,0 +1,139 @@
+# Proxy contracts
+
+* Author: anon_real
+* Status: Proposed
+* Created: 01-Sep-2021
+* License: CC0
+* Forking: not needed 
+
+## Motivation 
+Decentralized auctioning of any kind of tokens (artwork, share tokens, etc.) is an important part of any blockchain. This EIP is proposing the auction contract with various features listed in the [Design](#design) section
+
+
+## Design
+This proposed contract allows any kind of tokens to be auctioned while supporting the following features:
+- Any token as the auction's currency alongside ERG
+- "Buy it now" which allows a buyer to instantly buy the auctioned token by paying the predefined amount in the auction's currency
+- Auction duration as timestamp which is much more precise than block height and is independent of the network difficulty
+- Auto extending the duration near the end of the auction based on a global config
+- Minimum bid step - each bidder has to increase the previous bid at least by this amount
+- Royalty - The original owner (whomever issued the token) gets a share of the auction every time his/her token is auctioned using a global config
+
+## The contract
+```scala
+{
+  // R4: The seller's ergo tree
+  // R5: Current bidder's ergo tree
+  // R6: Minimum bid step
+  // R7: The auction's end time in timestamp
+  // R8: The auction's "Buy it now" amount. -1 if it is not enabled.
+  // R9: Auction info that is needed for UI/UX purposes - comma separated string encoded as Coll[Byte]:
+  //     - The auction's initial bid
+  //     - The auction's start time
+  //     - The auction's description
+  //     - Any other info needed in the future
+  //
+  // tokens(0): auctioned token
+  // tokens(1): current bid for non-ERG auctions - empty otherwise
+  //
+  // dataInputs(0): Auction house configuration - contains configs like UI fee and artist fee
+  //     - R4: UI fee, e.g. 100 for 1% or 200 for 0.5%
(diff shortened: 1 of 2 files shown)


greenhat:
I'm testing the contract at https://github.com/ergoplatform/sigma-rust/pull/390 and so far I found that we are missing `box.bytes` op. I'm going to implement it and see if anything else is missing (fail on deserialization). So far I've glanced through the contract and I think everything else we have implemented already.

greenhat:
I implemented `box.bytes` and finished checking the contract. It should work now.  
I merged my PR - https://github.com/ergoplatform/sigma-rust/pull/390

eip-0022.md
```diff
@@ -0,0 +1,169 @@
...
+     currBid < minBid &&
+     CONTEXT.preHeader.timestamp >= endTime &&
+     OUTPUTS.size == 2 &&
+     OUTPUTS(0).tokens(0) == SELF.tokens(0) &&
+     OUTPUTS(0).propositionBytes == seller &&
+     getBoxVal(OUTPUTS(0)) >= currBid - minerFeeInCurrency
+  }
+
+  val validConfig = auctionConfig.tokens(0)._1 == AUCTION_CONFIG_TOKEN_ID
+  sigmaProp((isNextBid || isFinishedWithBid || isFinishedWithoutBid) && validConfig && 2 > 1)
```
kushti: Why 2>1 part ?

kushti (approved):

eip-0022.md
anon-real: Removed!

chriswill:
I had an interation with @anon-real on Discord about this line:
// R9: Auction info that is needed for UI/UX purposes - comma separated string encoded as Coll[Byte]:
and he agreed that it would make more sense to represent this as json so we could have a collection of name/value pairs. 

Just wanted to add this comment so that ideally the EIP could be updated before it was approved and merged.

I think this is important for NFT developers would want to create a collection of custom properties and later display them.  Would also be value for displaying nfts in auctionhouse.

For an example, see the existing NFT id:3124c2d48a3ce19f69d92f71eb4cf0ab41cfec2515f636d06fdd916a22838581

anon-real:
> I had an interation with @anon-real on Discord about this line: // R9: Auction info that is needed for UI/UX purposes - comma separated string encoded as Coll[Byte]: and he agreed that it would make more sense to represent this as json so we could have a collection of name/value pairs.
> 
> Just wanted to add this comment so that ideally the EIP could be updated before it was approved and merged.
> 
> I think this is important for NFT developers would want to create a collection of custom properties and later display them. Would also be value for displaying nfts in auctionhouse.
> 
> For an example, see the existing NFT id:3124c2d48a3ce19f69d92f71eb4cf0ab41cfec2515f636d06fdd916a22838581

@chriswill I think you are mixing two separate things. R9 register for auction box has some information about the auction such as description. This is different from NFT data which is set when the NFT is issued.
Anyway, json format makes more sense -- I've updated the PR to include this change.

eip-0022.md
```diff
@@ -0,0 +1,182 @@
...
+  // either auction has ended due to time or "Buy it now" is being used
+  val isFinishedWithBid = {
+     val winnerBox = OUTPUTS(0)
+     val auctionFeeBox = OUTPUTS(1)
+     val sellerBox = if (OUTPUTS.size < 3) SELF
+                         else OUTPUTS(2)
+     val artistRoyalty = if (OUTPUTS.size < 4) SELF
+                         else OUTPUTS(3)
+     val originalIssuanceBox = if (auctionFeeBox.R4[Box].isDefined) auctionFeeBox.R4[Box].get
+                               else OUTPUTS(OUTPUTS.size - 1)
```
greenhat, line 127: Since the original issuance box cannot be in outputs I guess the reason to get an arbitrary box from the outputs is to have it fail later on id check, right? Would having `val originalIssuanceBox = auctionFeeBox.R4[Box]` at option be more readable?

greenhat (approved):
Looking good. See my minor comment.

eip-0022.md
anon-real: Yes, that would be better. However, this is a trick to get around how the costing currently works. Since both `if` and `else` parts will be evaluated (sometimes, it's weird) that wouldn't work and would produce exceptions.

kushti (approved):

PR #41, opened 2021-09-07
EIP-0023 Oracle pool 2.0

scalahub:
Main features in v2.0

1. Single pool address (only one type of box). This version of the pool will have only one address, the *pool address*. 
2. Data-point boxes will be considered as inputs rather than data-inputs. These inputs will be spent, and a copy of the box with the reward will be created.
3. Reward in tokens not Ergs. The posting reward will be in the form of tokens, which can be redeemed separately. 
4. Reward accumulated in data point boxes to prevent dust. We will not be creating a new box for rewarding each posting. Instead, the rewards will be accumulated in data-point boxes.
5. When creating a data-point box, the pool box is not needed as data input. Creating a data-point will be decoupled from the pool, and will not require the bool box as data-input. 
6. Update mechanism as before. We will have the same update mechanism as in v1.0
7. Transferable participant tokens. Participant tokens are free to be transferred between public keys
8. Longer epoch period (1 hour or 30 blocks).
9. No separate funding box. The pool box emits only reward tokens and won't be handing out Ergs. Thus, there won't be a separate funding process required. 
10. Reward mechanism separate from pool. The process of redeeming the reward tokens is not part of the protocol.


scalahub:
https://github.com/ergoplatform/eips/blob/eip23/eip-0023.md#reward-mechanism

Document mentions "We also give a sample token exchange contract", which seems to be missing. Either add the contract or remove the sentence

eip-0023.md
```diff
@@ -0,0 +1,530 @@
...
+| 0 | Oracle | Oracle |  
+
+This is used to transfer the ownership of an oracle token to another public key.
+
+The transaction is similar to the [publish data-point](#publish-data-point) transaction, except that Step 3 is modified as follows.
+
+3. The following rules (not enforced by the smart contract) to be followed.
+   - It should store the new owner's group element in R4.
+   - It should keep at least 1 reward token.
+   - It should store the balance reward tokens, if any, in some other box.
```
greenhat: This effectively makes this operation 2-in-1 (oracle token transfer + extract reward tokens). I suggest we leave the reward token extraction out and focus on oracle token transfer. 
We can either preserve the reward tokens in the output box and effectively transfer them to the new owner or run this operation only if there is only one reward token in the box. I prefer the second option to avoid accidental rewards token transfers.

scalahub: yes, we could allow ownership transfer only when 1 reward token is present. The 2-in-1 was by design, but accidental token transfer can become an issue.

greenhat: Please, update the description.

eip-0023.md
```diff
@@ -0,0 +1,530 @@
...
+|Refresh-NFT | Refresh box <br /> Reward tokens | Update-NFT <br /> Update box  <br /> Ballot tokens <br /> Ballot boxes <br /> Oracle tokens <br /> Oracle boxes (with new reward token) |
+|Update-NFT | Update box <br /> Ballot tokens <br /> Ballot boxes  | Refresh-NFT <br /> Refresh box <br /> Reward tokens <br /> Oracle tokens <br /> Oracle boxes |
+|Oracle tokens | Refresh-NFT <br /> Refresh box <br /> Reward tokens <br /> Oracle boxes <br /> | Update-NFT <br /> Update box <br /> Ballot tokens <br /> Ballot boxes |
+|Reward tokens | Refresh-NFT <br /> Refresh box | Update-NFT <br /> Update box <br /> Ballot tokens <br /> Ballot boxes <br /> Oracle tokens <br /> Oracle boxes |
+|Ballot tokens | Update-NFT <br /> Update box <br /> Ballot boxes | Refresh-NFT <br /> Refresh box <br /> Reward tokens <br /> Oracle tokens <br /> Oracle boxes |
+
+After an update, the older versions of the boxes will be left for garbage collection via storage-rent, and their tokens will become free for miners to take.
```
greenhat: I don't understand how to update a refresh box contract (to change epoch length, etc.). Should we mint a new Refresh-NFT and make a new refresh box, update pool box contract via voting ballots, mint new reward tokens, etc.? It looks like starting a new oracle pool should be easier.

scalahub: Yes new refresh NFT and new refresh box. Refer screenshot.
New oracle may be easier but not always possible because oracleNFT is hardwired in dApps

Screenshot from 2022-05-11 21-51-14

greenhat: Sure, I saw the table. I just found it's incomplete for the changes in the refresh contract.
Let's see what we need to do in order to change the epoch length:
- mint new refresh nft token;
- update pool contract/box (refresh nft is hard-coded the pool contract);
- mint new reward tokens;
- **force oracle boxes to use the new reward token. How? I did not find a way yet;**
- make a new refresh box;

As you can see there is one show-stopper (oracle boxes) and all these steps need to be done while oracles are running and can spend the old refresh box at any time.

scalahub: To change reward token oracle can spend the box using private key. See the test here https://github.com/scalahub/OraclePool/blob/v2/src/test/scala/oraclepool/v2/OracleSpec.scala#L39

Also search for "reward token" in this page for related tests, such as creating a box without any reward tokens 

https://github.com/scalahub/OraclePool/blob/v2/src/test/scala/oraclepool/v2/OracleSpec.scala#L46

There is a typo in the comment though (reward tokens can be 0 or more, instead of 1 or more)

Is it an issue if the oracle updates it while running? Each oracle can update reward token separately (reward token is sent to oracle in a simple box or maybe it can be sent in a special box that allows it to be loaded only into an oracle box)

greenhat: Delegating the switch to a new reward token for the oracles is certainly the way. I just wanted to find a way to keep the whole "change epoch length" operation local, i.e. run by one party. Which in hindsight kind of contradicts the decentralized notion of the oracle pool. 

So, let's iterate on the workflow needed to change the epoch length:
1. Mint the new refresh nft;
2. Run voting via ballots and update the pool contract/box;
3. Mint new reward tokens;
4. Send one new reward token to each oracle (to be put in the new oracle box in the next step);
5. Coordinate with **every** oracle to run a command to create a new oracle box and transfer oracle nft for each oracle;
6. Make a new refresh box;

I find this workflow quite complex. Also, issuing new reward tokens bothers me. I assume there will be some infrastructure built around reward tokens to make a profit for oracles. It should be updated as well in order to transition to the new reward token, right?  

Can we find a way to avoid issuing new reward tokens and updating oracle boxes in this case (updating pool parameters)?

greenhat: Could you please update this PR with changes from v2.0a from https://github.com/scalahub/OraclePool/tree/v2/src/main/scala/oraclepool/v2a

scalahub: Have updated contracts. Please review again

greenhat (approved):
Looking good! 
To be completely sure, I'd wait with PR merging until we implement voting and update operations.

eip-0023.md
```diff
@@ -0,0 +1,584 @@
...
+  val updateBoxOut = OUTPUTS(1) // copy of this box is the 2nd output
+
+  // compute the hash of the pool output box. This should be the value voted for
+  val poolOutHash = blake2b256(poolOut.propositionBytes)
+  val rewardTokenId = poolOut.tokens(1)._1
+  val rewardAmt = poolOut.tokens(1)._2
+  
+  val validPoolIn = poolIn.tokens(0)._1 == poolNFT
+  
+  val validPoolOut = poolIn.propositionBytes != poolOut.propositionBytes  && // script should not be preserved
```
greenhat: This requirement will get in the way if one wants to change only the reward token without changing the pool box contract. Can we alter/ditch it?

scalahub: We can probably remove the "script change requirement", unless it leads to some security issue (I cannot immediately see one).

greenhat: I agree. I don't see any security issue.

eip-0023.md
```diff
@@ -0,0 +1,584 @@
...
+               INPUTS(1).tokens(0)._1 == updateNFT       && // can only update when update box is the 2nd input
+               output.R4[GroupElement].get == selfPubKey && // public key is preserved
+               output.value >= SELF.value                && // value preserved or increased
+               ! (output.R5[Any].isDefined)                 // no more registers; prevents box from being reused as a valid vote 
+  
+  val owner = proveDlog(selfPubKey)
+  
+  // unlike in collection, here we don't require spender to be one of the ballot token holders
+  isSimpleCopy && (owner || update)
+}```
```
greenhat:
```suggestion
}
```
```
to fix markdown rendering

File: eip-0023.md (hunk 3 above)
Comment (scalahub) [reply eip-0023.md]: I've updated the contract to remove this check, and added v2c in the testing repo.

File: eip-0023.md [hunk 5]
```diff
@@ -0,0 +1,584 @@
+# Oracle Pool v2.0 
+
+* Author: @scalahub, @greenhat
```
greenhat:
```suggestion
* Author: @scalahub, @greenhat, @kettlebell, @SethDusek
```
Since most of my requests for updates lately actually came from @kettlebell and @SethDusek.

eip-0023.md
```diff
@@ -0,0 +1,584 @@
...
+  
+  val validUpdateOut = updateBoxOut.tokens == SELF.tokens                     &&
+                       updateBoxOut.propositionBytes == SELF.propositionBytes &&
+                       updateBoxOut.value >= SELF.value                       &&
+                       updateBoxOut.creationInfo._1 > SELF.creationInfo._1    &&
+                       ! (updateBoxOut.R4[Any].isDefined) 
+
+  def isValidBallot(b:Box) = if (b.tokens.size > 0) {
+    b.tokens(0)._1 == ballotTokenId       &&
+    b.R5[Int].get == SELF.creationInfo._1 && // ensure vote corresponds to this box by checking creation height
```
greenhat: @scalahub I cannot understand the reason behind this check. I mean, why require the ballots and check the creation height of the existing update box? Why pool hash check below is not enough to make sure ballots are voting for this change?

sethdusek: I think one possible case might be a revert update.

For example, a minimum of 6 votes might be needed to update pool. The first update U1 gets 12 votes, and only 6 of the ballots were spent & reset during update. A while later, another update U2 is performed. It might still be possible for someone to take the ballots that voted for U1 and revert the pool to U1

greenhat: Oh, I see it now.  Makes sense. Thank you!

greenhat:
@scalahub Could you please merge https://github.com/ergoplatform/eips/pull/78 into this PR?

eip-0023.md
```diff
@@ -0,0 +1,584 @@
...
+
+  // compute the hash of the pool output box. This should be the value voted for
+  val poolOutHash = blake2b256(poolOut.propositionBytes)
+  val rewardTokenId = poolOut.tokens(1)._1
+  val rewardAmt = poolOut.tokens(1)._2
+  
+  val validPoolIn = poolIn.tokens(0)._1 == poolNFT
+  
+  val validPoolOut = poolIn.tokens(0) == poolOut.tokens(0)                && // NFT preserved
+                     poolIn.creationInfo._1 == poolOut.creationInfo._1    && // creation height preserved
```
greenhat: @scalahub After EIP-39 (creation height of the output cannot be less than max input box creation height) implemented in https://github.com/ergoplatform/ergo/pull/1865 this requirement (pool box creation height preservation) is no longer satisfiable. We need to come up with a new requirement. What was the reason for this requirement in the first place?

scalahub: @greenhat I have removed the creation height preserving condition. Please review. Note that I've not verified that the individual contracts are same as the one in the single file.

greenhat: Looking good! Thank you! I'm going to deploy a new pool and test the update.
I verified that the contracts in extracted files are identical.

eip-0023/eip-0023.md
```diff
@@ -0,0 +1,337 @@
...
+
+Before going into the transactions in the protocol, we first present the contracts for each of the above boxes.
+
+## Contracts with base-64-encoded hash of ergo-tree bytes
+
+- [Pool Contract](contracts/pool_contract.es) `8cJi+FGGU32jXyO8M2LeyWSWlerdcb1zxBWeZtyy7Y8=`
+- [Refresh Contract](contracts/refresh_contract.es) `cs5c5QEirstI4ZlTyrbTjlPwWYHRW+QsedtpyOSBnH4=`
+- [Oracle Contract](contracts/oracle_contract.es) `fhOYLO3s+NJCqTQDWUz0E+ffy2T1VG7ZnhSFs0RP948=`
+- [Ballot Contract](contracts/ballot_contract.es) `2DnK+72bh+TxviNk8XfuYzLKtuF5jnqUJOzimt30NvI=`
+- [Update Contract](contracts/update_contract.es) `0wFmk/1TNpgTsbzpWND3WLPbwQdD8E+TWDzZLaYv3nE=`
```
greenhat:
```suggestion
- [Update Contract](contracts/update_contract.es) `3aIVTP5tRgCZHxaT3ZFw3XubRV5DJi0rKeo9bKVHlVw=`
```
Generated with https://scastie.scala-lang.org/BxJFCRDcSCeiwf9ZgKdymQ
using the latest update contract https://github.com/ergoplatform/eips/blob/1757e0060d6dafa6bd9bcb2339ab5484dfc767a7/eip-0023/contracts/update_contract.es with pool box preservation requirement removed.
Corresponding PR in oracle-core is https://github.com/ergoplatform/oracle-core/pull/178

eip-0023/eip-0023.md
```diff
@@ -0,0 +1,337 @@
...
+Before going into the transactions in the protocol, we first present the contracts for each of the above boxes.
+
+## Contracts with base-64-encoded hash of ergo-tree bytes
+
+- [Pool Contract](contracts/pool_contract.es) `8cJi+FGGU32jXyO8M2LeyWSWlerdcb1zxBWeZtyy7Y8=`
+- [Refresh Contract](contracts/refresh_contract.es) `cs5c5QEirstI4ZlTyrbTjlPwWYHRW+QsedtpyOSBnH4=`
+- [Oracle Contract](contracts/oracle_contract.es) `fhOYLO3s+NJCqTQDWUz0E+ffy2T1VG7ZnhSFs0RP948=`
+- [Ballot Contract](contracts/ballot_contract.es) `2DnK+72bh+TxviNk8XfuYzLKtuF5jnqUJOzimt30NvI=`
+- [Update Contract](contracts/update_contract.es) `3aIVTP5tRgCZHxaT3ZFw3XubRV5DJi0rKeo9bKVHlVw=`
+Use this Scastie playground to calculate the above hashes - [https://scastie.scala-lang.org/hnTEm2lJQPG1wRYCqPz8LQ](https://scastie.scala-lang.org/hnTEm2lJQPG1wRYCqPz8LQ)
```
greenhat, line 147: formatting

```suggestion

Use this Scastie playground to calculate the above hashes - [https://scastie.scala-lang.org/hnTEm2lJQPG1wRYCqPz8LQ](https://scastie.scala-lang.org/hnTEm2lJQPG1wRYCqPz8LQ)
```

eip-0023/contracts/update_contract.es
```diff
@@ -0,0 +1,55 @@
...
+                       updateBoxOut.value >= SELF.value                       &&
+                       updateBoxOut.creationInfo._1 > SELF.creationInfo._1    &&
+                       ! (updateBoxOut.R4[Any].isDefined) 
+
+  def isValidBallot(b:Box) = if (b.tokens.size > 0) {
+    b.tokens(0)._1 == ballotTokenId       &&
+    b.R5[Int].get == SELF.creationInfo._1 && // ensure vote corresponds to this box by checking creation height
+    b.R6[Coll[Byte]].get == poolOutHash   && // check proposition voted for
+    b.R7[Coll[Byte]].get == rewardTokenId && // check rewardTokenId voted for
+    b.R8[Long].get == rewardAmt              // check rewardTokenAmt voted for
```
greenhat, line 47: Handled in the suggestion above.
```suggestion
```

eip-0023/contracts/ballot_contract.es
```diff
@@ -0,0 +1,33 @@
...
+               INPUTS(1).tokens.size > 0                 &&
+               INPUTS(1).tokens(0)._1 == updateNFT       && // can only update when update box is the 2nd input
+               output.R4[GroupElement].get == selfPubKey && // public key is preserved
+               output.value >= SELF.value                && // value preserved or increased
+               ! (output.R5[Any].isDefined)                 // no more registers; prevents box from being reused as a valid vote 
+  
+  val owner = proveDlog(selfPubKey)
+  
+  // unlike in collection, here we don't require spender to be one of the ballot token holders
+  isSimpleCopy && (owner || update)
```
greenhat: @scalahub `isSimpleCopy` clause requires minting new ballot tokens for ANY change to the update contract or its parameters (min_votes). Do you see that it's really necessary here? Is not PK owner check not enough? I cannot think of a scenario where PK check is not enough. 

I forgot to mint new ballot tokens on changing the update contract and now those ballot tokens cannot be used. See https://github.com/ergoplatform/oracle-core/issues/291#issuecomment-1609089590 for details.

Rewriting the final clause as `owner || (isSimpleCopy && update)` would allow to keep using existing ballot tokens since voting is done by ballot owners (PK check).

scalahub: It should be fine as per a quick review.

greenhat: Great! Thanks! I'll make a PR and bootstrap a new pool on testnet.

greenhat: I make PR for this change https://github.com/ergoplatform/eips/pull/94

eip-0023/eip-0023.md
```diff
@@ -0,0 +1,337 @@
...
+| 0 | Oracle | Oracle |  
+
+This is used to transfer the ownership of an oracle token to another public key.
+
+The transaction is similar to the [publish data-point](#publish-data-point) transaction, except that Step 3 is modified as follows.
+
+3. The following rules (not enforced by the smart contract) to be followed.
+   - It should store the new owner's group element in R4.
+   - It should keep at least 1 reward token.
+   - It should store the balance reward tokens, if any, in some other box.
```
greenhat: Following our discussion in https://github.com/ergoplatform/eips/pull/41#discussion_r870469418
```suggestion
   - It should check that there is only 1 reward token is in the oracle box and send it to the new owner along with oracle token.
```

greenhat (approved):
Looks good! I just added the fix to the oracle token transfer process. It's already implemented in the code.

eip-0023.md
greenhat: To be fixed in https://github.com/ergoplatform/eips/pull/41#discussion_r1291081690

PR #43, opened 2021-11-02
Artwork EIP

anon-real:
This EIP is an extension of asset standard and tends to standardize artwork issuance.

### DIFF
--- a/eip-0024.md
+++ b/eip-0024.md
@@ -1,4 +1,4 @@
-# Auction contract
+# Artwork contract
 
 * Author: anon_real
 * Status: Proposed
--- a/eip-0024.md
+++ b/eip-0024.md
@@ -21,3 +21,10 @@ The issuer box is important because it has the same ID as the artwork's token ID
 
 - R4 of this box is an Int, showing 1000 * royalty percentage of the artwork. e.g, 20 for 2% royalty. If R4 of the issuer box of an artowrk is empty or non-positive, then royalty is considered to be 0%.
 - Proposition bytes of this box is the contract that the royalty percentage will be sent to. In case of a simple proxy contract this means that the artist will recieve the royalty share. However, the proxy contract may inforce the royalty to go to any other complex contract - e.g., 20% to the artrist, 80% to some charity.
+
+## Artist Identity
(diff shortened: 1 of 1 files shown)


eip-0024.md
```diff
@@ -0,0 +1,23 @@
...
+## Design
+There are two important boxes in the issuance process of an artwork.
+- The issuance box, which follows [Asset standard](eip-0004.md) depending on the type of artwork.
+- The issuer box, which is the first input of the transaction that is issuing the artwork. In particular, the box with the same ID as the artwork's token ID.
+
+Now we will discuss why the issuer box is important and define a standard for it.
+
+The issuer box is important because it has the same ID as the artwork's token ID and we can include it (although it is spent) in our transactions either in registers or as context data. This means that we can prove/verify certain features that an artwork may have using this box. The following is the current standard for this box - which may be extended in the future.
+
+- R4 of this box is an Int, showing 1000 * royalty percentage of the artwork. e.g, 20 for 2% royalty. If R4 of the issuer box of an artowrk is empty or non-positive, then royalty is considered to be 0%.
```
aslesarenko, line 22: maybe use base points, like in finance, so that 20, means 0.2%.

eip-0024.md
```diff
@@ -0,0 +1,23 @@
+# Auction contract
```
aslesarenko:
```suggestion
# Artwork contract
```

eip-0024.md
anon-real: Ah, I wish I knew this before. Changing this requires some tricky changes in the AH code as well which I'd like to avoid if possible. If you think that should be done, I'll do it for sure though.

kushti (approved):

eip-0024.md
aslesarenko: AH app definitely serves are a reference implementation for this EIP. At the same time, once accepted it may be used in other apps, at which point it is better this app doesn't have such issues.
So I can approve the EIP, but leaving it for you to decide about this issue.

aslesarenko (approved):
Looks good, except one place (see comment).

PR #44, opened 2021-11-07
EIP-0018 ErgoFund

code-for-uss:


kushti (approved):
Approved, but please move old version into an Appendix, to have one file only

eip-0018.md
```diff
@@ -0,0 +1,387 @@
...
+|Campaign participant tokens | 100 million per campaign | Identify campaign and each pledge box| Capmaign box <br /> pledge boxes |
+
+## Boxes
+
+There are a total of 3 contracts, and each corresponds to a box-type + control box.
+
+| Box | Quantity | Tokens | Additional registers used | Purpose | Spending transactions |
+|---|---|---|---|---|---|
+|Control | 1 | ControlBoxNFT | R4: Price (Long), <br /> R5: address[script] (SigmaProp) | Control price and address to pay for campaign registration | <div align="center">-</div> | 
+|Sell| 1 | SellNFT <br /> Registration tokens | <div align="center">-</div> | Register a new campaign | registerCampaign, <br/> returnRegistrationToken |
```
anon-real, line 37: Why would you need to return the registration token since this box is already a bottleneck for starting crowdfunding? I guess you could burn the token instead or send it to the `false` script.

eip-0018.md
```diff
@@ -0,0 +1,387 @@
...
+|Campaign | 0 - 100 million | Registration token <br /> Campaign participant tokens | R4: Campaign description (Coll[Byte]), <br /> R5: Public key of project owner (SigmaProp), <br /> R6: Campaign deadline (Int), <br /> R7: minToRaise (Long), <br /> R8 (optional): campaign Proposal CID (Coll[Byte]) | Store campaign data and collect funds | Donate to campaign, <br /> Withdraw funds, <br /> Refund donates, <br /> Return registrationToken |
+|pledge | 0 - 100 million (per campaign) | Campaign participant tokens | R4: Donate value (Long), <br /> R5: Public key of backer (SigmaProp) | Store donate value and Backer public key | Refund donate (fundraising failure) |
+
+## Contracts
+
+### Pledge Contract
+
+This box can only be spent if the first input box, which is the campaign box related to this pledge. Which means the campaign has failed.
+(Conditions: fund collected for the project is less than the defined minToRaise and the height of the network exceeds the fundRaisingDeadline), also, in this case, the fee of funds' returning will be deducted from the backer's fund.<br />
```
anon-real, line 46: It can be considered to let the campaign owner finish the campaign if funds are raised even before the deadline.

eip-0018.md
anon-real: I am opposed to the idea of having one sell box. There is no reason for that in this case (unlike cases like SigmaUSD bank box)
No reason to create a bottleneck when there is no gain.

eip-0018.md
```diff
@@ -0,0 +1,387 @@
...
+     val campaignBox = INPUTS(0)
+     val refundBox = OUTPUTS(1)
+     val fundraisingFailure = campaignBox.tokens(0)._1 == campaignRegistrationTokenId       && 
+                              campaignBox.tokens(1)._1 == campaignId                        && // verify campaign id
+                              refundBox.value >= donateValue + SELF.value - minFee          && // ensure full refund
+                              refundBox.propositionBytes == backerPubKey.propBytes // propositionBytes of refundBox must equal to backerPubKey
+     sigmaProp(fundraisingFailure)
+  }
+```
+### Campaign Contract
```
anon-real, line 70: I am strongly against the idea of having campaign tokens, at least in this way.
The point of crowdfunding is to be able to collect many possibly small funds. Now we are making the campaign box a bottleneck for donating to the campaign and I see no gain for that.

With the current design, there is no point in having a separate campaign box and pledge box. You can easily get all funds in the campaign box and give the equivalent amount of token to users for a later possible refund. I am proposing to do that since this also suffers from the issue I explained above -- I am making the point that donating should not go through the campaign box.

anon-real: I have previously proposed an idea to @kushti about how to have campaign IDs. So if the point of having a campaign token is that, there are simpler solutions to that.

"For the ErgoFund contracts, I think the campaign ID should be controlled by the sell contract otherwise someone can mess with the UI and potentially impersonate a campaign" in a chat with kushti.

greenhat: I concur @anon-real, moreover, I'm wondering if there is an option to design the whole system without the accumulation in the campaign box.
Of the top of my head:
The campaign starts with a box with height and target amount in the registers. The box value is the fee to the ErgoFund service. Box id is used as campaign id or via purchased campaign token bought from ErgoFund.
Backers create a donation box that either can be spent by the campaign owner upon height and target donation amount requirements are met or by the backers themselves if the campaign is failed.
ErgoFund service is funded from the campaign box value and optionally by taking a % from donations (enforced in donation box contract).

greenhat (changes requested):

PR #45, opened 2021-11-10
eip-0021 add token "Melange"

ebron99:
Added native token "Melange"


kushti (approved):

MrStahlfelge (changes requested):
You've entered "Ergo" as issuer which only makes sense for a token issued by core developers / EF / other officials. Please give your own website so we can check the token purpose.

aslesarenko:
Closing for now due to this acceptance policy "Before opening a PR to add your token to this list, ask yourself if your token is interesting for scammers."

It can be re-opened in future, once it is necessary.

PR #49, opened 2021-12-14
EIP-0025 Payment Request URI scheme

MrStahlfelge:
This EIP defines a payment request URI scheme to enable users to easily make payments by simply clicking links on webpages or in e-mails.

In difference to Ergo Pay (EIP-0020), this URI scheme does not contain a prepared transaction that the wallet should sign or discard. Instead, it contains 
the data that the wallet application should use to fill its payment form. The user can change the details when needed. The wallet will build the transaction, 
so this scheme is easier to handle on static websites or in e-mails.

### DIFF
--- a/eip-0025.md
+++ b/eip-0025.md
@@ -51,8 +51,12 @@ Examples
 
      ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx
 
-     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx&amount=1.0&description=Hard%20work
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?amount=1.0&description=Hard%20work
 requests 1 ERG for hard work
 
-     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx&token-03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04=5
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?token-03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04=5
 requests 5 SigUSD
+
+In the case of sending token only, a minimum amount of ERGs MUST also be sent to a new box (it is not possible to create boxes containing tokens only with zero ERGs).
+
+If the `amount` parameter is not specified, then the wallet MUST put a minimum amount of ERGs to the new box.
--- a/eip-0025.md
+++ b/eip-0025.md
@@ -21,7 +21,7 @@ so this scheme is easier to handle on static websites or in e-mails.
 Format
 ------
 
-    ergourn        = "ergoplatform:" ergoaddress [ "?" ergoparams ]
+    ergourn        = "ergo:" ergoaddress [ "?" ergoparams ]
     ergoaddress    = *base58
     ergoparams     = ergoparam [ "&" ergoparams ]
     ergoparam      = [ amountparam / labelparam / messageparam / tokenparam ]
--- a/eip-0025.md
+++ b/eip-0025.md
@@ -1,0 +1,58 @@
+EIP-0025: URI scheme for payment requests
+=========================================
+
+* Author: MrStahlfelge
+* Created: 14-Dec-2021
+* License: CC0
+* Forking: not needed
+
+Motivation
+----------
+
+Like [BIP-0021](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) for Bitcoin, this EIP proposes a URI scheme for initiating payments.
+
+The purpose of this URI scheme is to enable users to easily make payments by simply clicking links on webpages or in e-mails.
+
+In difference to Ergo Pay (EIP-0020), this URI scheme does not contain a prepared transaction that the wallet should sign or discard. Instead, it contains 
+the data that the wallet application should use to fill its payment form. The user can change the details when needed. The wallet will build the transaction, 
+so this scheme is easier to handle on static websites or in e-mails.
+
+
+Format
+------
+
+    ergourn        = "ergoplatform:" ergoaddress [ "?" ergoparams ]
+    ergoaddress    = *base58
+    ergoparams     = ergoparam [ "&" ergoparams ]
+    ergoparam      = [ amountparam / labelparam / messageparam / tokenparam ]
+    amountparam    = "amount=" *digit [ "." *digit ]
+    labelparam     = "label=" *qchar
+    messageparam   = "description=" *qchar
+    tokenparam     = "token-" qchar *qchar "=" *digit [ "." *digit ]
+
+
+* label: Label for that address (e.g. name of receiver)
+* address: ergo P2PK or P2S address
+* description: message that describes the transaction to the user
+ 
+### Amount
+
+If an amount is provided, it MUST be specified in decimal ERG. All amounts MUST contain no commas and use a period (.) as the separating character to separate 
+whole numbers and decimal fractions. I.e. amount=50.00 or amount=50 is treated as 50 ERG, and amount=50,000.00 is invalid.
+
+### Tokens
+
+If a token parameter is provided, it specifies the token id and the desired amount separated by an equals sign. The amount must be specified in decimal token 
+value, for example "3.45" for 3.45 SigUSD. The amount MUST contain no commas and use a period (.) as the separating character to separate whole numbers and decimal fractions.
+
+
+Examples
+--------
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx&amount=1.0&description=Hard%20work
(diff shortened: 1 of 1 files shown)


eip-0025.md
```diff
@@ -0,0 +1,58 @@
...
+If a token parameter is provided, it specifies the token id and the desired amount separated by an equals sign. The amount must be specified in decimal token 
+value, for example "3.45" for 3.45 SigUSD. The amount MUST contain no commas and use a period (.) as the separating character to separate whole numbers and decimal fractions.
+
+
+Examples
+--------
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx&amount=1.0&description=Hard%20work
```
aslesarenko:
```suggestion
     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?amount=1.0&description=Hard%20work
```

eip-0025.md
```diff
@@ -0,0 +1,58 @@
...
+requests 1 ERG for hard work
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx&token-03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04=5
```
aslesarenko:
```suggestion
     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?token-03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04=5
```

eip-0025.md
```diff
@@ -0,0 +1,58 @@
...
+requests 5 SigUSD
```
aslesarenko, line 58:
```suggestion
requests 5 SigUSD

In the case of sending token only, a minimum amount of ERGs MUST also be sent to a new box (it is not possible to create boxes containing tokens only with zero ERGs).

If the `amount` parameter is not specified, then the wallet MUST put a minimum amount of ERGs to the new box.
```

aslesarenko (approved):

satsen:
I believe that `ergo:` should be used instead of `ergoplatform:`. I know that ergo is a common word, which is why ergoplatform.org had to be used instead of `ergo.org`, but I highly doubt that the URI scheme name is used by anything else. Imagine using URLs like `bitcoinplatform:` or `moneroplatform:`, it's not nice at all.

satsen:
I'm also unsure about the `If the amount parameter is not specified, then the wallet MUST put a minimum amount of ERGs to the new box` part. Being that loose is of no use and anybody who has sent tokens will already know this. Inexperienced users could also be confused if the wallet improperly represents that it is about to send some ERGs as well when the parameter is missing.

aslesarenko:
@satsen Some ERGs have to be sent anyway, this is required by the protocol. So, what is your proposal, make ERGs amount a required parameter?
Please suggest a formulation.

satsen:
@aslesarenko No, it should not be required. For for example donations, it doesn't make sense to require the parameter. My proposal is that if a token amount is supplied, then the ERG amount is required.

eip-0025.md
```diff
@@ -0,0 +1,62 @@
...
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?amount=1.0&description=Hard%20work
+requests 1 ERG for hard work
+
+     ergoplatform:9ewA9T53dy5qvAkcR5jVCtbaDW2XgWzbLPs5H4uCJJavmA4fzDx?token-03faf2cb329f2e90d6d23b58d91bbb6c046aa143261cc21f52fbe2824bfcbf04=5
+requests 5 SigUSD
+
+In the case of sending token only, a minimum amount of ERGs MUST also be sent to a new box (it is not possible to create boxes containing tokens only with zero ERGs).
+
+If the `amount` parameter is not specified, then the wallet MUST put a minimum amount of ERGs to the new box.
```
aslesarenko, line 62:
```suggestion
In the case of sending token only, a minimum amount of ERGs MUST also be sent to a new box (it is not possible to create boxes containing tokens only with zero ERGs). Thus, if a token amount is supplied, then the ERG amount parameter is required.
```

aslesarenko (changes requested):

MrStahlfelge:
How should a wallet app react when a token is added in the URI, but no amount?
It looks like the whole payment request should be treated as invalid and should be discarded then? This does not seem very user friendly, I would prefer to let the wallet app set the min amount automatically in that case. This behaviour was requested by users for manual added tokens as well, most were confused that they had to specify an ERG amount manually.

eip-0025.md
MrStahlfelge: Because of the reasons mentioned in my comment, I would prefer Thus, if a token amount is supplied, then the ERG amount parameter is recommended to use.

MrStahlfelge:
> I believe that `ergo:` should be used instead of `ergoplatform:`. I know that ergo is a common word, which is why ergoplatform.org had to be used instead of `ergo.org`, but I highly doubt that the URI scheme name is used by anything else. Imagine using URLs like `bitcoinplatform:` or `moneroplatform:`, it's not nice at all.

I think ergo is a bit too generic, and there is already a ergopay URI scheme prefix proposed. Let's see what the majority thinks or prefers here.

satsen:
@MrStahlfelge You're not really going to generate ergo URIs that much for the average user to need this kind of looseness. A wallet program could either leave the amount field empty and require the user to enter an ERG amount or it could deny the URI (not good). I think the standard should be that ERG amount is required when there is a token amount, but of course not everyone will follow the standard and asking the user for the missing amount should be the method in my opinion.

satsen:
@MrStahlfelge I don't know what ergopay is or why a project that is not ergo should affect what the URI scheme name will be.

satsen:
Do you mind specifying what happens if there are 2 entries of the same token. I suspect and propose that the last one will be the valid one but it would be good to have it in the standard.

aslesarenko:
@MrStahlfelge I think the wallet should process the URI as long as it can parse and interpret the content unambiguously.
Any ambiguity should be detected and if it cannot be resolved, the wallet should give a descriptive message to the user (not just "Invalid URI", but what exactly has been detected as invalid).

At the same time the wallet should not do anything implicitly, even if it is smart enough to so some implicit actions, it should display what is going on to the user on the transaction sending screen. The user should be aware that some implicit action where taken by wallet and that those actions will take effect when the Send button is pressed.

Applying the above principles to the missing ERGs amount parameter problem we can see that it can be automatically resolved by the wallet, so it shouldn't reject the URI as invalid when the amount of ERGs is missing.
At the same time, the screen should have an explanation that non-zero amount of ERGs in the field is **required** and that it was suggested by the wallet.

For the case of duplicate tokens, I think it is better to consider such URI as invalid, and clearly communicate the problem to the user so that he can take a screenshot and send it back to the dApp developer. The duplication problem should be clear from the screenshot.

eip-0025.md
aslesarenko:
```suggestion
The wallet should process the URI as long as it can parse and interpret the content unambiguously. Any ambiguity should be detected and if it cannot be resolved, the wallet should give a descriptive message to the user (not just "Invalid URI", but what exactly has been detected as invalid).

At the same time the wallet should not do anything implicitly, even if it is smart enough to so some implicit actions, it should display what is going on to the user on the transaction sending screen. The user should be aware that some implicit action where taken by wallet and that those actions will take effect when the Send button is pressed.

Applying the above principles to the missing ERGs amount parameter problem we can see that it can be automatically resolved by the wallet, so the wallet shouldn't reject the URI as invalid when the amount of ERGs is missing. At the same time, the screen should have an explanation that non-zero amount of ERGs in the field is required and that it was suggested by the wallet.

For the case of duplicate tokens, the wallet CAN consider such URI as invalid, and clearly communicate the problem to the user so that the user can take a screenshot and send it back to the dApp developer. The duplication problem should be clear from the screenshot.
```

aslesarenko (changes requested):
I suggest to clarify the principles in the EIP text

MrStahlfelge:
> @MrStahlfelge I don't know what ergopay is or why a project that is not ergo should affect what the URI scheme name will be.

[See here](https://github.com/ergoplatform/eips/pull/37)

aslesarenko (approved):

greenhat (approved):

eip-0025.md
```diff
@@ -0,0 +1,68 @@
...
+
+In difference to Ergo Pay (EIP-0020), this URI scheme does not contain a prepared transaction that the wallet should sign or discard. Instead, it contains 
+the data that the wallet application should use to fill its payment form. The user can change the details when needed. The wallet will build the transaction, 
+so this scheme is easier to handle on static websites or in e-mails.
+
+
+Format
+------
+
+    ergourn        = "ergoplatform:" ergoaddress [ "?" ergoparams ]
```
satsen: It should be changed to "ergo:" only. To follow other cryptocurrencies (bitcoin, monero, etc) and because URI schemes need to be short, for example telegram calls its scheme for tg. It's not generic as the name of the crypto is commonly the payment request. If only using the name is bad, then the URI should be ergoplatformpaymentrequest, which is ridiculous.

satsen (changes requested):

eip-0025.md
aslesarenko: I aggree, there are merits to using "ergo". 
However if we consider other possible schemes "ergopay", "ergomixer", "ergoteam", "ergodex" etc, in which "ergo" is followed by some application name, then "ergoplatform" will make sense also, as well as "ergopaymentrequest".

kushti: ergoplatform is too verbose to me too

kushti (approved):

PR #50, opened 2021-12-18
EIP-5: Added TreeVersion property and other updates

aslesarenko:
We now have version 0 and 1 of ErgoTree on chain. 
With upcoming v5.0 and then v6.0 we will have even more versions. 
ErgoTree is interpreted according to the tree.version bits in its header. 
Thus, producing ErgoTree from contract template should respect the version, if it is specified.

### DIFF
--- a/eip-0005.md
+++ b/eip-0005.md
@@ -51,19 +51,20 @@ here only for description of the corresponding content. Thus, the format is sche
 Binary serialization format is suitable for compact storage and wire transfers. All the
 fields are required (mandatory). 
 
-Field Name     |  Format            | Example                | Description
----------------|--------------------|------------------------|-----------
-`NameLength`   | `VLQ(UInt)`        | 18                     | Length of `Name` bytes (> 0)
-`Name`         | `Bytes`            | "rewardOutputScript"   | User readable name (non-empty string bytes in UTF-8 encoding)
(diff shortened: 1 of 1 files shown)


eip-0005.md
```diff
@@ -78,15 +79,18 @@ serialized as Data format prefixed by 0 - if the value is `None`, and 1 - if the
...
 are not segregated. In most cases this is more compact and can be done by the compilers
 (e.g. [ErgoScala compiler](#notes-on-supporting-this-eip-in-ergotree-compilers)) where
 constant <-> parameter mapping is explicit. For such contracts the whole `ConstValues` field
 may be serialized as None.
 
-- `TemplateSize` field can be used to skip parsing `Template` when the contract template
-is embedded into some another format. Thus templates can be quickly scanned and the
-`Template` bytes obtained for hashing or template-based requests to the blockchain.
+- `TemplateSize` field can be used to skip parsing `ExpressionTree` when the contract template
+is embedded into some another format. Thus, templates can be quickly scanned and the
```
oskin1:
```suggestion
is embedded into some other format. Thus, templates can be quickly scanned and the
```

oskin1 (approved):

greenhat (approved):

PR #53, opened 2022-01-25
EIP-0028 ErgoAuth

MrStahlfelge:
ErgoAuth: user authentication protocol between wallet applications and dApps

### DIFF
--- a/eip-0028.md
+++ b/eip-0028.md
@@ -9,7 +9,7 @@
 ## Contents
 - [Description](#description)
 - [Background And Motivation](#background-and-motivation)
-- [ErgoPay Interaction Protocol](#ergoauth-authentication-protocol)
+- [ErgoAuth authentication protocol](#ergoauth-authentication-protocol)
 - [Data Formats](#data-formats)
 - [Implementation in Mobile Wallet](#implementation-in-mobile-wallet)
 - [Implementation in dApp](#implementation-in-dapp)
@@ -24,7 +24,7 @@ dApp.
 
 dApps might want to validate if dApp users are really who they pretend to be. This is especially
 useful for dApps that grant certain abilities to holders of special tokens. At the moment,
-proofing that a user owns a token can only be done by sending the token to a depositary address.
+proving that a user owns a token can only be done by sending the token to a depositary address.
 By sending the token, the user proofed to have access privileges to the token.
 However, sending token around is not always desirable. Especially for valuable tokens, users might
 not want to send it away, and doing two transactions (one to send it to the depositary address, one
--- a/eip-0028.md
+++ b/eip-0028.md
@@ -97,7 +97,7 @@ ErgoAuthRequest:
 (Remark: An Ergo p2pk address is a SigmaBoolean, so authenticating a wallet address is possible with this)
 
 The **signingMessage** is a String that is not user-friendly to read in general, as it might contain
-information the dApp adds to make it unique. If the signingMessage containsa 0-byte character (unicode 0000), 
+information the dApp adds to make it unique. If the signingMessage contains 0-byte character (unicode 0000), 
 the part of the signingMessage befire this sign is interpreted as the user prompt what he is going to sign for
 and must be shown to the user.
 
--- a/eip-0028.md
+++ b/eip-0028.md
@@ -1,0 +1,137 @@
+# ErgoAuth: user authentication protocol between wallet applications and dApps
+
+* Author: @MrStahlfelge
+* Status: Proposed
+* Created: 25-Jan-2022
+* License: CC0
+* Forking: not needed
+
+## Contents
+- [Description](#description)
+- [Background And Motivation](#background-and-motivation)
+- [ErgoPay Interaction Protocol](#ergoauth-authentication-protocol)
+- [Data Formats](#data-formats)
+- [Implementation in Mobile Wallet](#implementation-in-mobile-wallet)
+- [Implementation in dApp](#implementation-in-dapp)
+- [Benefits for dApps](#benefits-for-dapps)
+- [Benefits for Mobile Wallets](#benefits-for-mobile-wallets)
+
+## Description
+This EIP defines a standard for trustless authentication of users of a wallet app and an online
+dApp.
+
+## Background And Motivation
+
+dApps might want to validate if dApp users are really who they pretend to be. This is especially
+useful for dApps that grant certain abilities to holders of special tokens. At the moment,
+proofing that a user owns a token can only be done by sending the token to a depositary address.
+By sending the token, the user proofed to have access privileges to the token.
+However, sending token around is not always desirable. Especially for valuable tokens, users might
+not want to send it away, and doing two transactions (one to send it to the depositary address, one
+to refund it back) costs both time and transaction fees.
+
+To overcome this, ErgoAuth proposes a way to authenticate users trustless to have access to certain
+addresses storing a box. The protocol is trustless in both ways: The users don't need to trust
+the dApp, because the dApp does not get access to funds or secrets. The dApp don't need to trust
+the users or the wallet app, because it can validate the authentication keys.
+
+## ErgoAuth authentication protocol
+
+An authentication with ErgoAuth is driven by a dApp that needs to authenticate a user.
+
+1) The user enters the necessary information in the dApp's UI for the dApp to know if authentication
+is necessary. For example, users might enter their P2PK address (or, instead of manually entering,
+use ErgoPay to send the address to the dApp automatically).
+
+2) The dApp determines that authenticating the user is needed. For this, the dApp prepares a unique
+message that the wallet app should sign with a user's private key, and an Ergo address that the user
+needs to authenticate for. This might be a P2PK address, but in general P2S addresses can be also
+used for authentication.
+
+3) The dApp presents an ErgoAuth link for the user to click and open the wallet app and a QR code
+for mobile users to scan from within the wallet app.
+
+4) The wallet application parses the QR code/link data and obtains a
+`ErgoAuthRequestUrl` to fetch the actual `ErgoAuthRequest` data from
+(see [Data Formats](#data-formats) section).
+
+5) When `ErgoAuthRequest` is obtained, the wallet presents a screen showing that a dApp wants to
+authenticate the user, and the address the request is for. The wallet app should also inform the
+user that no funds or moved and no secrets will leave the device.
+
+6) When the user agrees, the wallet app signs the obtained message from ErgoAuthRequest and sends
+the signed message to the ErgoAuthRequest's replyToUrl.
+
+7) The dApp validates the signed message. When successful, it can proceed with its flow.
+
+## Data Formats
+
+Wallet apps should be able to initiate ErgoAuth both by using URI schemes
+(clickable links) or QR codes.
+
+`ergoauth://<URL>`
+
+An URL is provided without the https prefix. http communication is not allowed except for IP addresses
+(in order to test within a local network).
+
+Examples:
+* `ergoauth://sigmavalley.io/auth/2021-16b8-66c4-b800-6e52-8ce4` will make the wallet app request
+`https://sigmausd.io/auth/2021-16b8-66c4-b800-6e52-8ce4`
+* `ergoauth://192.168.0.1/auth` will make the wallet app request
+`http://192.168.0.1/auth`
+
+#### Response body: ErgoAuthRequest
+
+The wallet application should request URL and obtain the following data (json format)
+
+```
+ErgoAuthRequest:
+  - signingMessage: String
+  - address: String
+  - userMessage: String (optional*)
+  - messageSeverity: String (optional) "INFORMATION", "WARNING"
+  - replyToUrl: String
+```
+
+If provided, the wallet application should show the **userMessage** and display the **messageSeverity**
+in a suitable way.
+
+After signing is performed and the `SignedMessage` data is obtained, the
+wallet must POST the following data to the dApp using **replyToUrl** from the
+request (json format).
+
+```
+ErgoAuthResponse:
+  - signedMessage: String (base64)
+```
+
+In case there was an error building the ErgoAuthRequest on the dApp side, the dApp might reply
+with an `ErgoAuthRequestError` to inform the user about the error:
+
+```
+ErgoAuthRequestError:
+  - userMessage: String
+```
+
+The wallet application will show the user message to the user.
+
+
+## Implementation in wallet app
+TODO
+
+## Implementation in dApp
+TODO
+
+## Benefits for dApps
+- A dApp or website don’t need to handle user's secrets (mnemonic/private keys), but can
+safely validate if a user has access to certain Ergo addresses.
+- dApp's users don't need to worry about security of their private keys as the
+wallet guarantees they never leave the device
+
+
+## Benefits for Wallet apps
+
+Any wallet team should consider supporting ErgoAuth in their wallet along
+with basic wallet features. Users can participate in Ergo dApps and mobile
+wallet team can receive service fees from those transactions.
+
--- a/eip-0028.md
+++ b/eip-0028.md
@@ -94,6 +94,8 @@ ErgoAuthRequest:
   - replyToUrl: String
 ```
 
+(Remark: An Ergo p2pk address is a SigmaBoolean, so authenticating a wallet address is possible with this)
+
 If provided, the wallet application should show the **userMessage** and display the **messageSeverity**
 in a suitable way.
 
@@ -128,7 +130,10 @@ The wallet application will show the user message to the user.
 TODO
 
 ## Implementation in dApp
-TODO
+
+Ergo Appkit: ErgoAuthUtils helper class adds helper methods for server side applications. ErgoAuthSpec demonstrates the use.
+
+dApp: TODO
 
 ## Benefits for dApps
 - A dApp or website don’t need to handle user's secrets (mnemonic/private keys), but can
--- a/eip-0028.md
+++ b/eip-0028.md
@@ -44,9 +44,8 @@ is necessary. For example, users might enter their P2PK address (or, instead of 
 use ErgoPay to send the address to the dApp automatically).
 
 2) The dApp determines that authenticating the user is needed. For this, the dApp prepares a unique
-message that the wallet app should sign with a user's private key, and an Ergo address that the user
-needs to authenticate for. This might be a P2PK address, but in general P2S addresses can be also
-used for authentication.
+message that the wallet app should sign with a user's private key, and a SigmaBoolean that the user
+needs to authenticate for. This might be a P2PK address wrapped in a SigmaBoolean.
 
 3) The dApp presents an ErgoAuth link for the user to click and open the wallet app and a QR code
 for mobile users to scan from within the wallet app.
@@ -58,6 +57,8 @@ for mobile users to scan from within the wallet app.
 5) When `ErgoAuthRequest` is obtained, the wallet presents a screen showing that a dApp wants to
 authenticate the user, and the address the request is for. The wallet app should also inform the
 user that no funds or moved and no secrets will leave the device.
+In a future enhancement, the Auth Request could be relayed to a Cold wallet device. This is an enhancement 
+of EIP-0019 and would not change ErgoAuth protocol.
 
 6) When the user agrees, the wallet app adds some own bytes to the obtained message from ErgoAuthRequest,
 signs it  and sends the signed message to the ErgoAuthRequest's replyToUrl.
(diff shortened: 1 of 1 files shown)


eip-0028.md
```diff
@@ -0,0 +1,137 @@
...
+* Author: @MrStahlfelge
+* Status: Proposed
+* Created: 25-Jan-2022
+* License: CC0
+* Forking: not needed
+
+## Contents
+- [Description](#description)
+- [Background And Motivation](#background-and-motivation)
+- [ErgoPay Interaction Protocol](#ergoauth-authentication-protocol)
```
aslesarenko:
```suggestion
- [ErgoAuth authentication protocol](#ergoauth-authentication-protocol)
```

eip-0028.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Description
+This EIP defines a standard for trustless authentication of users of a wallet app and an online
+dApp.
+
+## Background And Motivation
+
+dApps might want to validate if dApp users are really who they pretend to be. This is especially
+useful for dApps that grant certain abilities to holders of special tokens. At the moment,
+proofing that a user owns a token can only be done by sending the token to a depositary address.
```
aslesarenko:
```suggestion
proving that a user owns a token can only be done by sending the token to a depositary address.
```

eip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+`http://192.168.0.1/auth`
+
+#### Response body: ErgoAuthRequest
+
+The wallet application should request URL and obtain the following data (json format)
+
+```
+ErgoAuthRequest:
+  - signingMessage: String
+  - address: String
```
aslesarenko: this should not be Address, as in general Address may require evaluation. This instead should be serialized SigmaBoolean.

eip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+  - userMessage: String (optional*)
+  - messageSeverity: String (optional) "INFORMATION", "WARNING"
+  - replyToUrl: String
+```
+
+If provided, the wallet application should show the **userMessage** and display the **messageSeverity**
+in a suitable way.
+
+After signing is performed and the `SignedMessage` data is obtained, the
```
aslesarenko: SignedMessage is not specified.

eip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+(see [Data Formats](#data-formats) section).
+
+5) When `ErgoAuthRequest` is obtained, the wallet presents a screen showing that a dApp wants to
+authenticate the user, and the address the request is for. The wallet app should also inform the
+user that no funds or moved and no secrets will leave the device.
+
+6) When the user agrees, the wallet app signs the obtained message from ErgoAuthRequest and sends
+the signed message to the ErgoAuthRequest's replyToUrl.
+
+7) The dApp validates the signed message. When successful, it can proceed with its flow.
```
aslesarenko: An extension of the workflow may be specified which uses Cold Wallet for signing.

eip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+TODO
+
+## Implementation in dApp
+TODO
+
+## Benefits for dApps
+- A dApp or website don’t need to handle user's secrets (mnemonic/private keys), but can
+safely validate if a user has access to certain Ergo addresses.
+- dApp's users don't need to worry about security of their private keys as the
+wallet guarantees they never leave the device
```
aslesarenko: Mention integration with Cold Wallet (and potentially with Hardware wallet)

eip-0028.md
MrStahlfelge: I had the return value from `ErgoProvierImpl.signMessage` in mind. I don't know how to reference it any better, do you have a suggestion?

eip-0028.md
MrStahlfelge: I think in most of the cases, a dApp wants to authenticate a holder of a certain box or owner of a P2PK address. So to keep it simple for the dApps, my plan was to use the contract address here from which the ErgoTree can be extracted. From the ErgoTree, we can evaluate to the SigmaBoolean.

eip-0028.md
MrStahlfelge: For the moment, cold wallets won't work with ErgoAuth because the interchange format between a cold and a hot wallet can only transfer transaction signing requests for now. I think instead of extending it here, we need to extend the Cold Wallet EIP later.

eip-0028.md
aslesarenko: We can evaluate ErgoTree to SigmaBoolean only in the presence of Context data. This brings us back to ReducedTransaction as a way to avoid having the Context on the wallet. While Appkit fully supports ErgoTree evaluation, other wallet app platforms may not. 

For this reason, you cannot reduce an arbitrary address to SigmaBoolean for further signing on such incapable platforms. 

But, if a dApp takes care of Address to SigmaBoolean reduction, then it can support the case you are describing and also present plain simple SigmaBoolean to the wallet for signing.

eip-0028.md
aslesarenko: the signMessage method returns `Array[Byte]` of the proof (aka signature), and it is better encapsulate it into some named type (like `Signature`). 
The `SignedMessage` is also good, but similar to SignedInput it should have both the original message and the proof bytes.
Something like: 
```
SignedMessage:
  - message: Array[Byte]
  - proofBytes: Array[Byte]
```

eip-0028.md
aslesarenko: Yes, here you can refer to the next Cold Wallet EIP revision (mentioning it is not yet implemented).
Will serve as an advertisement for this EIP.

eip-0028.md
MrStahlfelge: Okay, when we want to use `SigmaBoolean`, this makes it more easy on the wallet side. To keep it simple for dApps that only want to verify p2pk addresses, we could use both ways (like there are two methods on appkit implemented).
But: is there already a serialization format for `SigmaBoolean` that is implemented in both appkit and sigma-rust?

aslesarenko: Yes, in Sigma it is https://github.com/ScorexFoundation/sigmastate-interpreter/blob/985ce41e2e44d82a44ac57b700c7c735f855428e/sigmastate/src/main/scala/sigmastate/Values.scala#L799
and I belive something similar exists in sigma-rust, because SigmaBoolean is part of the protocol.

aslesarenko (approved):

aslesarenko (approved):

aslesarenko (approved):

eip-0028.md
```diff
@@ -0,0 +1,144 @@
...
+* `ergoauth://192.168.0.1/auth` will make the wallet app request
+`http://192.168.0.1/auth`
+
+#### Response body: ErgoAuthRequest
+
+The wallet application should request URL and obtain the following data (json format)
+
+```
+ErgoAuthRequest:
+  - signingMessage: String
```
capt-nemo429, line 94: I think we can simplify the API even more by removing the need of `signingMessage`, as the protocol requires the wallet to add some random bytes, it will be a random message anyway.

In the end, the protocol is about proving that you own a specific private key, so I see no need to ask for a message for that.

Alternatively, we can make this optional, if a dev wants additional entropy `signingMessage` can be used for it.

MrStahlfelge: If the signingMessage is removed, there is no proof that the user signed a message for you (the initiating dApp) any more. Then it is enough for a bad actor to get knowledge of a single signed message to reeuse it for authentication to all dApps. Thus it is needed and must be mandatory.

anon-br: Oh, you are right, simple sk ownership proof can be dangerous, I didn't think about this possibility.

capt-nemo429: Oh, you are right, simple sk ownership proof can be dangerous, I didn't think about this possibility.

eip-0028.md
```diff
@@ -0,0 +1,144 @@
...
+## ErgoAuth authentication protocol
+
+An authentication with ErgoAuth is driven by a dApp that needs to authenticate a user.
+
+1) The user enters the necessary information in the dApp's UI for the dApp to know if authentication
+is necessary. For example, users might enter their P2PK address (or, instead of manually entering,
+use ErgoPay to send the address to the dApp automatically).
+
+2) The dApp determines that authenticating the user is needed. For this, the dApp prepares a unique
+message that the wallet app should sign with a user's private key, and a SigmaBoolean that the user
```
nirvanush, line 48: Possible vulnerability. If I understand correctly the link has to be generated on BE. In order to do that Client need to make a request. Here is a scenario:
1. dApp A (Good dapp) and dApp B (Bad actor) implement the same authentication.
2. Client want to sign to dApp B.
3. dApp B makes a request to dApp A to get a unique message.
4. dApp B ask client to sign that message. and client signs/submits it to dApp B
5. dApp B submit the signature to dApp A in exchange for an authentication token, now dAppB's dev have an access to Client's dApp A protected routes.

eip-0028.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Background And Motivation
+
+dApps might want to validate if dApp users are really who they pretend to be. This is especially
+useful for dApps that grant certain abilities to holders of special tokens. At the moment,
+proving that a user owns a token can only be done by sending the token to a depositary address.
+By sending the token, the user proofed to have access privileges to the token.
+However, sending token around is not always desirable. Especially for valuable tokens, users might
+not want to send it away, and doing two transactions (one to send it to the depositary address, one
+to refund it back) costs both time and transaction fees.
```
nirvanush, line 32: User can just send a token to himself. And there is no need to submit that transaction either, just to check that the transaction is valid via transaction/check endpoint.

eip-0028.md
MrStahlfelge: Not really a vulnerability, depending on context it could be wanted, but it could also be unwanted. 
I see following possibility to work against an unwanted case:
* Wallet application adds the hostname of the replyto endpoint to the signed message. This way, a dApp can check if the message was directly sent or relayed.

capt-nemo429: Having the hostname added to `signedMessage` sounds like a really good idea, a timestamp can be also a good addition.

## `signedMessage` pattern suggestions:
We can enforce a pattern that can be simple to parse and verify, for this a semicolon delimited string seems like a good choice, a JSON formatted string is an alternative, a bit verbose one but even simpler to parse.

### Proposed fields
 * `origin`: domain, subdomains and TLD
 * `timestamp`: Unix timestamp at the moment of the signature. Useful for dApps to guarantee message validity periods.

### Semicolon delimited string
* Pattern: `"signingMessage;origin;timestamp;randomBytes"`
* Example `"someRandomString;www.sigmavalley.org;1656538465;b955cba40ceb"`

### JSON formatted string
* Pattern: `"{signingMessage:string,origin:string,timestamp:string,randomBytes:string}"`
* Example: `'{"signingMessage":"someRandomString","origin":"www.sigmavalley.org","timestamp":"1656538465","randomBytes":"b955cba40ceb"}'`

MrStahlfelge: I think semicolon separation is okay here. To make it more random, a prefix could be used:
signingMessageRandomByteso:originRandomBytes
so dApp can check if o:(own host name) is present.

Timestamp does not hurt to have, but in my opinion it is not really necessary since the creation of the signingMessage can be persisted on dApp's side.

capt-nemo429: Oh, great! Not sure about the prefix tho. We can make it more random by just increasing the `randomBytes` length, I'm using an UUID here.

MrStahlfelge: Just embedding without prefix works, too. It is more important to define what part of URL or hostname should be used.

Wallet App adds random bytes before and after the signing message. This is not mandatory but allowed, just wanted to state it here.

nirvanush: > * Wallet application adds the hostname of the replyto endpoint to the signed message. This way, a dApp can check if the message was directly sent or relayed.
Doesn't help. Bad actor can submit the signed message using headless chrome, tools like Puppeteer. In that case origin will match

MrStahlfelge: I don't know Puppeteer but from what I see it looks like you are suggesting that the bad actor has access to the user's system. If this is the case, no guarantees exist. If not, please elaborate.

nirvanush: > I don't know Puppeteer but from what I see it looks like you are suggesting that the bad actor has access to the user's system. If this is the case, no guarantees exist. If not, please elaborate.

Puppeteer is a package over Google Chrome binary that is used for web scraping and integration tests for web apps.
It allows me to imitate browser behavior with code, fill forms, and scrap pages. I can do a request with Puppeteer and it will have the same origin as user's request. No need to access the user's system.

**So just for an order, Let's start from the beginning.**

**The problem that we are trying to solve is to allow user<>server authentication.**

In traditional apps it's done with a **password**. ServerA  stores passwordA and ServerB stores passwordB of the same user. Technically user can use the same pass for ServerA and B, but it is considered a bad practice because if passA gets exploited the hacker will get an access to all apps the user uses. 
The core principle of that authentication is that apps don't share passwords between them.

In the ErgoAuth, the offered solution is that server will have two routes
- `serverA/getMessageToSign`
- `serverA/submitSignedMessage`

User will have to receive a message from the server by sending HTTP request to that server to the first route.
The problem with that is if this route is public, anyone can send a request to it. Here is a possible scenario:
image

In this scenario BadApp instead of sending user it's own unique message, takes a message from GoodApp and let user sign it. Now BadApp can submit the signed message to GoodApp and gain access to user's data on GoodApp.


The solution that you guys have provided suggests checking origin of the request to verify that the request was sent by user and not another server.

**Why it is not good.**
1. Origin is just an HTTP header and can be easily set/edited by anyone it's not built to be used as request security purposes. CORS policies are enforced by browsers, not by servers.
 Usually, if you want to protect a communication between servers it'd done with an internal token or VPC. All servers inside VPC are allowed to communicate. But you can't do it with public routes, because you want any client be able to access it.

2. Origin is basically a page where the browser is located. 
image
Hacker can easily fake it by request /serverB/getMessageToSign with Puppeteer session: Locate Puppeteer on AppB page and make a request.


In the current solution there is no secret on the server it's a problem needs to be solved..

@MrStahlfelge @capt-nemo429

MrStahlfelge: I think cause for the misunderstanding is our use of the word origin. When talking about the origin, I referred to the replyToUrl and I guess @capt-nemo429 was also doing so.
I know this can also be compromised by a man in the middle attack, but it is way harder to do. If your services are subject to be attacked by someone doing this type of effort, you should probably work with other authentication mechanisms.

aslesarenko: @MrStahlfelge 
The messages between a user and a server should probably be encrypted using pk/sk pairs.
The user can know pk of the server and use it for encryption.
The server can encrypt its replies to the user with user's pk, so that only user can decrypt it (with his sk).
In this case no man in the middle attack.

MrStahlfelge: I need some more elaboration here to understand that approach completely.

However, I am not sure if we should just lower the expectations for ErgoAuth. ErgoAuth should not be used for usages with (financial) disadvantages of the user. It is meant to be used to grant access to restricted areas of a website/game/app and to proove ownership. The only risks involved should be that the application grants access to too many users, which is no real (financial) damage. And when it is, or if any spendings or other (financial) disadvantages arise for the user, a normal transaction or a second factor should be used.

MrStahlfelge: I have added the hostname to the proposal.

capt-nemo429: I think you forgot about the timestamp ;)

MrStahlfelge: No:

> Timestamp does not hurt to have, but in my opinion it is not really necessary since the creation of the signingMessage can be persisted on dApp's side.

Better dApp saves the timestamp, so it knows when it asked for the authentication. After all, we can have malicious players not only in the middle, but also on client side. EDIT: But of course that doesn't mean we can't add it. I just see more difficulties here for dApps in finding the timestamp within the signed message, while the hostname is pretty easy to find with a simple contains()

capt-nemo429: The timestamp is just a way to check message expiration time, so it is best to be added by the wallet since it can add it at the signing time.

If the dApp manages it, a user may take while on the signing process and end up by signing an already expired message.

MrStahlfelge: That is usually how it works, many 2FA mechanisms give you the request to authenticate and expire if you don't do it in time. Anway, not opponing it, but maybe others have opinions as well.

capt-nemo429: Yeah, they have the option of not using it. So It's better to have it than not. )

MrStahlfelge: True, my fear is that the ones using it are not aware of the implications

capt-nemo429: > True, my fear is that the ones using it are not aware of the implications

I don't see any problem on having it honestly.

> EDIT: But of course that doesn't mean we can't add it. I just see more difficulties here for dApps in finding the timestamp within the signed message, while the hostname is pretty easy to find with a simple contains()

If we use the pattern I suggested above, a dev can check it as simple as `signedMessage.split(";")[2]`

nirvanush: @MrStahlfelge @aslesarenko I'm not talking about the man in a middle attack, I'm talking about a much simpler case of dapp A propagating users' requests to server B to get access. Anyone can do this no hacking skills are needed and encryption will not help here. User thinks that he is signing a message of A, while in reality, it's a message of B. Once A has a signed message from the user it can submit it to B, now A have an access to B.

nirvanush: @MrStahlfelge @capt-nemo429 , what really can solve it is a simple solution. And let me know if this is what you are guys were talking about the whole time. :D

* The message must include the domain (origin)

* **Wallets (dapp connector) should prohibit from user to sign message if the `message.split(':')[0] !== <window.origin>` exactly how CORS policies  on browser side works (or maybe just set a huge warning like in burn token case?)**
image

* Server should validate that the message includes the origin. 

Without the wallet part, it will not work. If wallets enforces the message standard AppA will not be able to substitute the message with AppB's message. And again if this is what you were talking about all the time, sorry for the misunderstanding!

nirvanush: But then again @MrStahlfelge why you want the server to generate that message and not the wallet itself?
I think server should be an option and if message is not provided the wallet could generate it using timestamp:origin:randombytes pattern or whatever

aslesarenko (changes requested):
I think ErgoAuth should be based on asymmetric encryption of messages between dApp (which have pk/sk key pair) and wallet users (which also have at least one pk/sk key pair).

If all the messages of the described ErgoAuth protocol got encrypted with a public key of the receiver, then the whole protocol will be safe and the messages can be sent via public channels.
When dApp sends ErgoAuthRequest, it should be encrypted with user's pk. (so only owner of user sk can read it)
Similarly, when user sends ErgoAuthResponse is should be encrypted with dApp's pk (so only dAppA can read it)

MrStahlfelge:
Maybe we should step back and go to the root cause of the problem. We discuss at the moment a problem that a man in the middle getting knowledge of a signed message can use it to authenticate to another dapp. Why do we have this problem here, but not in signed transactions in the first place? Because for transactions, the user get revealed what he actually is going to sign, while for now this is not the case here for ErgoAuth. The message that is signed is hidden to the user and only the issueing dapp knows what it is used for. To overcome this, we can define that the message to sign must contain a plain text message to be shown to the user, that explains in simple language what he is going to sign for. This plain text message could be prefixed to the message to sign and separated by a 0 byte.
Some examples for such a message: 

"Login to getblok.io configuration @ 06 Jul 22 15:30"
"Flip a coin @ NightOwl"
"Draw a card @ Blitz TCG"

etc etc

By this, the user knows what he is going to sign and we can get rid of the unsolvable problem of insecure transmission channels. Additionally, dapps can persist the signed message and signature and have a proof that the user actually signed to do an outlined action.

eip-0028.md
MrStahlfelge: Let's take discussion to the main PR thread

aslesarenko:
As an alternative to implementing encryption as part of this EIP, we can restrict this EIP to work only over secure https connections, also requiring the UI to show connection security status to the user (same as it is done in browsers)

MrStahlfelge:
> As an alternative to implementing encryption as part of this EIP, we can restrict this EIP to work only over secure https connections, also requiring the UI to show connection security status to the user (same as it is done in browsers)

This is now part of the specification:

To fetch the auth request:
> An URL is provided without the https prefix. http communication is not allowed except for IP addresses (in order to test within a local network).

For the replyToUrl:
> The replyToUrl's hostname must be the same as the one the request was fetched from - a wallet application should verify that.

So https is already enforced, I will check possibilities for showing the certificate information.

MrStahlfelge:
I have checked how FIDO works and it has the same concept to prevent middleman attacks. So I am confident we are good to go with this one. On a side note, ZelID only uses short living signing messages. This is recommended to do for the dApps anyway.

aslesarenko (approved):

PR #58, opened 2022-02-18
EIP-29 Attachment

MrStahlfelge:


eip-0029.md
```diff
@@ -0,0 +1,67 @@
...
+Ergo boxes can hold Coll[Byte] that can be a UTF-8 representation of the purpose field.
+However, this would mean that we must designate a certain register to be used as purpose register,
+and because of the restrictions mentioned above this would mean to use R4, blocking it for all
+other use cases. This is not desired.
+
+To overcome this, we define that the purpose is encoded as follows in an arbitrary box:
+
+`Pair([0x50, 0x52, 0x50], Coll[Byte])`
+
+This is a pair of a constant Coll[Byte] used to identify this register as a purpose message
```
JimmyHoffa: How about "any" register as a purpose message? Any reason we can't *allow* people to put a memo running through all additional registers?

eip-0029.md
```diff
@@ -0,0 +1,67 @@
...
+of `3c0e0e0350525`.
+
+The first register identified as a purpose message register can be used to read a purpose message
+for this box. If there is no such register on a box, no purpose message is defined.
+
+#### Examples
+
+| Purpose message         | Register hex representation                                        |
+| ----------------------- |:------------------------------------------------------------------:|
+| "Your loan January"     | 3c0e0e0350525011596f7572206c6f616e204a616e75617279                 |
```
JimmyHoffa: For sanitization, would it maybe be good to demand the data format encoded be something with known injection protection? Just thinking JSON parsers are known to safely parse JSON, XML parsers as well may, ProtoBuf as well etc.

When a receiving application tries to render the message, if it is a plain string and they don't purposefully sanitize it before rendering, it could present an injection attack vector on that application. If they are forced to use a known safe parser by making the message a specific format, it may create a safety net for applications consuming this field.

MrStahlfelge: There should be no injection possible with plain strings as long as they are not handled as something different. Using json serialization to handle a single string is just adding complexity as long as the serialized message is nothing different than a single string.

aslesarenko (approved):

aslesarenko (approved):

kushti (approved):

PR #59, opened 2022-02-24
Generalisation of idea of Purpose message

aslesarenko:
In this PR:
- Original idea is further generalized to support multiple types.
- Proposed technique to overcome the requirement of dense packing of registers (using Unit type)
- Register 9 is proposed for Attachment

### DIFF
--- a/eip-0029.md
+++ b/eip-0029.md
@@ -1,50 +1,112 @@
-# Transaction purpose field
+# Ergo Box Attachments
 
-* Author: MrStahlfelge
+* Author: MrStahlfelge, aslesarenko
 * Status: Proposed
+* Version: 1.0
 * Created: 18-Feb-2022
 * License: CC0
 * Forking: not needed 
 
 ## Motivation 
 
-Most payment systems provide a way to set a "purpose" field, used to display a short message
-from payment issuers to payment receivers.
+Historically, most payment systems provide a way to set a "purpose" field, used to display
+a short message from payment issuers to payment receivers. In the "state of the art" fintech
+practice sending money is converging with instant messaging platforms, where money can be
+sent among other messages i.e. directly in the chat.
 
-This EIP describes how one of the optional registers of Ergo boxes can be used to store a purpose
-message.
-This message can be set by wallet applications and dApps and can be shown by Ergo Explorer and
-Wallet applications.
+Ergo platform allows easily send assets stored in a _box_ in so that all network
+participants can read the contents of the box. However, there is no standard way to
+_attach_ additional data so that all participants can recognize and parse it.
 
+This EIP describes a data format for one of the optional registers of Ergo boxes that can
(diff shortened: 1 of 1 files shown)


eip-0029.md
```diff
@@ -1,50 +1,112 @@
...
 Registers must be densely packed. It is not possible to use register 9 without adding register 4
-to 8.
+to 8. However, if necessary registers of `Unit` type can be added to fill the empty slot
+with only 1 byte per register of overhead.
+
+A box cannot be more than 4 kbytes of serialized bytes.
+
+## Ergo Attachment register
+
+This standard specify that a particular register number 9 to be used as an Attachment
```
JimmyHoffa, line 49: Is there a reason to only use register 9?

I'd be concerned making the last register the **standard** may limit us to having a single attachment permanently, or result in people inventing alternative standard ways in the future.

If people implement to that standard  and then they ever want to make more attachments from their DApp, or an overflow standard allowing registers have continuation of the attachment, implementers will have to go back and change their implementation. Resulting in an implemented standard and this standard and confusion from there.

I suspect I'm just missing the reason not to make the magic bytes the indicator, could you maybe elaborate on why it needs to be magic bytes *and* register 9?

It seems we should be able to say any register after 4 with the magic bytes prefix is to be interpreted as an attachment and it allows us more features in the future.

Is the complexity of checking for magic bytes on all 4-9 registers the worry? This would be done off-chain so I feel it shouldn't be a concern?

aslesarenko: Those are very good questions.
1) Having a dedicated register to be used have some benefits:
- easy to access from ErgoScript, I foresee attachments will be used from contracts in dApp-specific way
- Easy to check existence of attachments from ErgoScript
- Provide clear separation of contract relevant from contract irrelevant registers
- Reduces the change of magic bytes to give false positive if by chance the contract's register have the same binary prefix in serialized bytes
2)  Content type = 1 solves the problem of multiple attachments, in fact the format allows to have a tree of attachments, and it is good to know all the attachments are in the same place
3) Typically contracts use a couple of registers in its logic and they usually R4, R5 and R6 i.e. densely packed. Sometimes the logic may depend on whether the register is present or not. Consider `if (SELF.R5[(Coll[Byte], (Int, Coll[Byte]))].isDefined)` condition in the contract, it refer to some data in the register. By chance the type of the data is the same as the type of attachment register, so if by chance an attachment is added to such box, it may interfere with contracts logic. That is why using the last register is better.

MrStahlfelge:
Like the more general idea. However, before merging into the original PR the examples must be updated (serialized string should differ with these suggestions)

aslesarenko:
@MrStahlfelge I suggest to update the examples at the end, based on the final version and the actual implementation.

MrStahlfelge:
Then let's have further discussion this in the original PR

PR #60, opened 2022-03-22
Update eip-0021.md

glasgowm148:
Adding `ADA` token


MrStahlfelge (approved):

MrStahlfelge:
We might also add "ADA" with unique name: yes and token id: empty to the verified tokens list. This way, all tokens named ADA would be marked suspicious.

arobsn (approved):

PR #62, opened 2022-04-19
EIP-30 Dexy stablecoin 

scalahub:


eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Emission Contract
+
+```scala
+{ 
+  // This box: (dexyUSD emission box)
+  //   tokens(0): emissionNFT identifying the box
+  //   tokens(1): dexyUSD tokens to be emitted
+  
+  val selfOutIndex = getVar[Int](0).get
```
kushti: Why not a constant?

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+  
+  val oraclePoolNFT = fromBase64("RytLYlBlU2hWbVlxM3Q2dzl6JEMmRilKQE1jUWZUalc=") // to identify oracle pool box
+  val swappingNFT = fromBase64("Fho6UlBlU2hWbVlxM3Q2dzl6JEMmRilKQE1jUWZUalc=") // to identify swapping box for future use
+  
+  val validEmission = {
+    val oraclePoolBox = CONTEXT.dataInputs(0) // oracle-pool (v1 and v2) box containing rate in R4
+  
+    val validOP = oraclePoolBox.tokens(0)._1 == oraclePoolNFT
+  
+    val oraclePoolRate = oraclePoolBox.R4[Long].get // can assume always > 0 (ref oracle pool contracts) NanoErgs per USD
```
kushti: per USD or USD cent?

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+  
+    val selfOut = OUTPUTS(selfOutIndex)
+  
+    val validSelfOut = selfOut.tokens(0) == SELF.tokens(0) && // emissionNFT and quantity preserved
```
kushti: "and quantity" could be missed

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+                     selfOut.propositionBytes == SELF.propositionBytes && // script preserved
+                     selfOut.tokens(1)._1 == SELF.tokens(1)._1 && // dexyUSD tokenId preserved
+                     selfOut.value > SELF.value // can only purchase dexyUSD, not sell it
```
kushti: "can only purchase dexyUSD, not sell it" => "can only sell dexyUSD, not purchase it", so
selfOut.value < SELF.value

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+  sigmaProp(validEmission || validTopping)
+}
+```
+## Liquidity Pool Contract
+
+```scala
+{
+    // Notation:
+    // 
+    // X is the primary token
```
kushti, line 271: X is the primary token (ERG)

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+    // Y is the secondary token 
+    // When using Erg-USD oracle v1, X is NanoErg and Y is USD
```
kushti, line 273: Y is USD cent ?

kushti (changes requested):

eip-0030.md
scalahub: Makes it more flexible due to possible AOTC issues with hard-wiring (box can be spent in two different transactions: top-up and mint)

eip-0030.md
scalahub: yes its USD-cent if we consider v1 oracle

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+
+    val reservesX0 = SELF.value
+    val reservesY0 = tokenY0._2
+    val reservesX1 = successor.value
+    val reservesY1 = tokenY1._2
+
+    val oraclePoolRateXY = oraclePoolBox.R4[Long].get 
+    val lpRateXY0 = reservesX0 / reservesY0  // we can assume that reservesY0 > 0 (since at least one token must exist) 
+    val lpRateXY1 = reservesX1 / reservesY1  // we can assume that reservesY1 > 0 (since at least one token must exist)
+    val isCrossing = (lpRateXY0 - oraclePoolRateXY) * (lpRateXY1 - oraclePoolRateXY) < 0 // if (and only if) oracle pool rate falls in between, then this will be negative
```
kushti: Why to check this in LP contract?

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+     
+    val crossCounterIn = SELF.R5[Int].get
+    val crossCounterOut = successor.R5[Int].get
+    
+    val validCrossCounter = crossCounterOut == {if (isCrossing) crossCounterIn + 1 else crossCounterIn}
```
kushti: Why to check this in LP contract?

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+        validCrossCounter
+    )
+}
+```
+## Tracking Contract
+
+```scala
+{
+  // Tracking box
+  //   R4: Crossing Counter of LP box
```
kushti: Why do we need for the counter? I guess it would be enough to write down in register height when price condition violated (within certain margin), and then a special action can nullify the register if price is okay again. On receiver side (in the bank contract), we check that HEIGHT - trackingBox.R4 > threshold

eip-0030.md
```diff
@@ -0,0 +1,329 @@
...
+      validLpBox &&
+      validOraclePoolBox && 
+      validThreshold &&
+      validSuccessor &&
+      validTracking
+     )
+}
+```
+
+## Swapping Contract
```
kushti: Description of the contract is needed.

kushti (changes requested):

eip-0030.md
scalahub: agreed, we can remove "and quantity"

eip-0030.md
scalahub: I believe the comment is misinterpreted. The full comment should read "users can only purchase dexyUSD". 

Maybe we can also write it as "box can only sell dexyUSD"?

Also since the contract sells dexyUSD, shouldn't selfOut.value be more than SELF.value?  (i.e., shouldn't Ergs increase?)

eip-0030.md
scalahub: I think better to keep X as nanoErgs

eip-0030.md
scalahub: yes comment is incorrect. It should be USD cent with oracle v1

eip-0030.md
scalahub: This variable captures whether the oracle pool rate falls in between the rates at input and output LP box in a swap transaction. Each time this happens, a counter is incremented in R5. This is done in every swap, independent of  the top-up requirement

eip-0030.md
scalahub: Just to keep track of how many times its crossed (see above comment). Note that the requirement we are trying to capture is that between a top-up initiation and top-up completion, the LP rate should not have crossed the oracle pool rate. If it does, then the top-up becomes invalid and we need to start afresh.

eip-0030.md
scalahub: Yes, there can be other ways to achieve the goal stated in previous comment. So in the LP, we would need to store two values. One contains boolean when condition is violated and another height at which violated. 

We can consider this approach as well. That way we can skip "initiation" and "completion" of top-up and just have a single step.

eip-0030.md
scalahub: Going to add in next commit

PR #65, opened 2022-05-10
EIP-4 add utility asset type

Luivatra:
Will allow for wallets and dapps to filter utility tokens and nft to make it easier for the user to keep an overview.


capt-nemo429:
I think that setting `R8` and `R9` just as "Optional" looks too vague and opens the door for unpredictable data format/content. I suggest making use of these records, we can even keep them as optional, but if any value is added, it should standardized.

### Examples:
* **LP token**: `R8 = baseToken, R9 = quoteToken`
* **Staking key**: `R8 = originalAmountStaked, R9 = stakeTime`
* **Wrapped asset**: `R8 = bridgeCode, R9 = (ticker, originChain)`

capt-nemo429:
I think that setting `R8` and `R9` just as "Optional" looks too vague and opens the door for unpredictable data format/content. I suggest making use of these records, we can even keep them as optional, but if any value is added, it should standardized.

### Examples:
* **LP token**: `R8 = baseToken, R9 = quoteToken`
* **Staking key**: `R8 = originalAmountStaked, R9 = stakeTime`
* **Wrapped asset**: `R8 = bridgeCode, R9 = (ticker, originChain)`

ghost:
I would like to see utilizing markup or superscript language as the standard for utility tokens, where layer two and wallets alike can parse something like yaml into arrangements of objects natively on chain

Luivatra:
I think having a sub-asset type for each dapp will (hopefully) become too much to list here. We could have R8 containing a string referring to a type, which a dapp is free to set themselves. R9 should be a JSON/yaml that has further description. fe:
R8: ErgoDex Liquidity Pool Token
R9: {'baseToken': 'erg', 'quoteToken': 'neta'}

Wallets and dapps can then always recognize utility tokens and parse R9 to show a "nice" description. If a specific use case arises for a certain type the wallet/dapp can check R8 for the type and validate R9 accordingly and fe. calculate a value for the LP token.

PR #66, opened 2022-05-10
eip27-edits: come corrections and todos

aslesarenko:


eip-0027.md
```diff
@@ -84,12 +84,6 @@ Contracts
 
     val correctNftId = EQ(firstTokenId, ByteArrayConstant(reemissionNftId))
     
-    // output of the reemission contract
```
kushti, line 87: why removed? reemissionOut and minerOut used below: 

```
    // miner's output must have a script which is the time-locking reward for the miner's pubkey
    // box height must be the same as block height
    val correctMinerOutput = AND(
      EQ(ExtractScriptBytes(minerOut), expectedMinerOutScriptBytesVal(monetarySettings.minerRewardDelay, MinerPubkey)),
      EQ(Height, boxCreationHeight(minerOut))
    )
    
    // reemission output's height must be the same as block height
    val heightCorrect = EQ(boxCreationHeight(reemissionOut), Height)
```

eip-0027.md
```diff
@@ -113,11 +107,15 @@ Contracts
     val correctCoinsIssued = EQ(reemissionRewardPerBlock, Minus(ExtractAmount(Self), ExtractAmount(reemissionOut)))
     
     // when reemission contract box got merged with other boxes
-    val sponsored = AND(
+    // TODO: describe why this condition can replace 5 others (heightCorrect, etc)
```
kushti, line 110: it is not replacing them, rather, an alternative branch. So re-emission box is whether getting additional funding via `sponsored` path, or spending funds according to rules (heightCorrect etc)

kushti (changes requested):

eip-0027.md
aslesarenko: But they are declare twice, see above.

eip-0027.md
aslesarenko: I don't understand what `sponsored` means, what is the intuition about this. Better to put it into the contract. 
Without understanding the intended logic, it is hard to review and estimate correctness of the contract.

kushti: Contract description in this EIP have this: 
```
Also, this contract allows for merging with other boxes; the merging transaction must have only two outputs; the first output is for the re-emission contract, the second one is to pay a mining fee (its value can be 0.01 ERG at most)
```

aslesarenko: Ok, I've read that. Why not also hint about what are the inputs of the merging transaction?

kushti: Well, there are no limits there, so re-emission contract box can be merged with any inputs allowing for outputs specified.

aslesarenko: Ok, the comment above inputs would be helpful in the code to decipher the contract.

PR #67, opened 2022-05-17
EIP-4 Introduce 'Asset Pack' NFT type

ghost:
Add new token type for "Asset Pack" which is proposed to be superscript language (YAML) referencing the ownership of other NFT objects in a hierarchy/pack. 
Allows for native organization on chain without layer two, easier organization on the layer two.

Enables things like music albums of multiple tracks, compilation works with multiple artists and pieces, in-game treasure chests containing other items, in-game houses containing other items, etc.


ghost:
Drawing

Drawing-2

Drawing-3

PR #69, opened 2022-06-05
EIP-34: Extension of artwork standard and proposing collection standard

anon-real:
The main goal of this pull request is as follows:
- Introducing collection standard
- Extending artwork standard to be both future-proof and support collections


eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+It must be noted that the actual value for traits will be present in the NFT data (refer to [artwork standard]()) and the above numbers for levels and stats only specify the maximum range.
+
+## Unlockable Content
+Unlockable content information must be present in `R8` of the issuance box as `Coll[Byte]`.
+Unlockable contents are files or other types of information that the artist reveals to the buyer when NFT is sold (and/or some other conditions are satisfied). At the time of proposing this EIP, there are no known decentralized solutions to have unlockable contents. However, it is more than likely that we'll have solutions for this soon so this part of this EIP introduces a structure that can be changed for backward and forward compatibility in the future.
+The following JSON contains the structure for unlockable contents.
+
+```json
+{
+    "version" standard_version, # this will change as new standards are needed to be added for unlockable contents
```
kii-dot: "version":

eip-0024.md
```diff
@@ -7,24 +7,50 @@
...
+
+- R4 of this box is an Integer, showing 1000 * royalty percentage of the artwork. e.g, 20 for 2% royalty. If R4 of the issuer box of an artowrk is empty or non-positive, then royalty is considered to be 0%.
+- In the current design of the [Auction contranct](eip-0022.md), proposition bytes of this box is the contract that the royalty percentage will be sent to. This is a suboptimal design because of AOT costing mechanism which is supposed to be replaced with JIT costing with a soft-fork that will happen with v5.0 of [ergo node](https://github.com/ergoplatform/ergo).
+  
+   In the case of a simple proxy contract, this means that the artist will receive the royalty share. However, the proxy contract may enforce the royalty to go to any other complex contract - e.g., 20% to the artist, and 80% to some charity.
```
kii-dot: Does this feel a little out of context? As in there's no context leading into this statement. (Starting from the In the case of a simple proxy contract)

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+    "logo": "URL_TO_COLLECTION_LOGO",
+    "featured_image": "URL_TO_COLLECTION_FEATURED_IMAGE",
+    "banner_image": "URL_TO_COLLECTION_BANNER_IMAGE",
+    "category": "COLLECTION_CATEGORY",
+    "explicit_content": true or false
+}
+```
+
+### Socials
+Socials must be present in `R6` of the issuance box as `Coll[Byte]`.
```
kii-dot: Must it be present? What if artists are not interested in sharing socials.

kii-dot (commented):
Would there be something similar to Cardano's policy id?

eip-0034.md
mgpai22: I agree, this should be optional.

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+
+Although levels and stats are very similar, they can be considered different in marketplaces both for UI/UX purposes and also for filtering purposes.
+
+The following specifies the structure for traits.
+```json
+{
+    "properties": ["NAME_OF_PROPERTY1", "NAME_OF_PROPERTY2"],
+    "levels": [
+        {
+            "NAME_OF_LEVEL1": max_alowed_level_number1
```
HazeyOneKenobi:
```suggestion
            "NAME_OF_LEVEL1": max_allowed_level_number1
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+        },
+        {
+            "NAME_OF_LEVEL2": max_alowed_level_number2
```
HazeyOneKenobi:
```suggestion
            "NAME_OF_LEVEL2": max_allowed_level_number2
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+        }
+    ],
+    "stats": [
+        {
+            "NAME_OF_STAT1": max_alowed_stat_number1
```
HazeyOneKenobi:
```suggestion
            "NAME_OF_STAT1": max_allowed_stat_number1
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+        },
+        {
+            "NAME_OF_STAT2": max_alowed_stat_number2
```
HazeyOneKenobi:
```suggestion
            "NAME_OF_STAT2": max_allowed_stat_number2
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+        },
+        {
+            "NAME_OF_STAT3": max_alowed_stat_number3
```
HazeyOneKenobi:
```suggestion
            "NAME_OF_STAT3": max_allowed_stat_number3
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+## Traits
+Traits information must be present in `R7` of the issuance box as `Coll[Byte]`.
+There are three types of traits:
+- Properties: These are textual traits such as specifying `sex` as `male`.
+- Levels: These are numerical traits that encode some sort of level-like information such as specifying `speed` as `60 out of 100`.
+- Stats: These are numerical traits that encode any numerical information about the NFT such as specifying `age` as `25 out of 50`.
+
+Although levels and stats are very similar, they can be considered different in marketplaces both for UI/UX purposes and also for filtering purposes.
+
+The following specifies the structure for traits.
```
HazeyOneKenobi: "specifies the structure for defining the traits that must be present in all NFTs of a collection" might be easier to understand

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+            "NAME_OF_STAT2": max_alowed_stat_number2
+        },
+        {
+            "NAME_OF_STAT3": max_alowed_stat_number3
+        }
+    ]
+}
+```
+
+The following is an example of traits for some imaginary NFT.
```
HazeyOneKenobi: "The following is an arbitrary example of the collection-token's trait data using the structure above. Traits of all NFTs minted under the collection must match with traits defined for the collection here"

eip-0034.md
HazeyOneKenobi: should we be hard-coding the social platform? Might be better to allow an artist to supply their socials / collection website as an array without keys, i.e. 

```json
{
	"socials": [
		"www.thisissplatform1.com/iMakeNfts",
		"www.thisissplatform2.com/iMakeNfts",
		"www.thisissplatform3.com/iMakeNfts"
	]
}
```

this would be a blank array if the artist chose to provide no social links.

eip-0034.md
```diff
@@ -0,0 +1,121 @@
+# NFT Collection Standard V1
+
+* Author: anon_real
+* Status: Proposed
+* Created: 10-May-2022
+* License: CC0
+* Forking: not needed 
+
+## Motivation 
+NFT collections play a central role in the nowadays NFT ecosystem on all blockchains. It gives the ability to artists to organize their artworks and present them in a more meaningful way. It also can facilitate the use of other NFT features such as traits. As we'll present in this proposal, it can also enable more advanced features such as artist collaborations. These advanced features can get arbitrary complex with the use of smart contracts.
```
HazeyOneKenobi:
```suggestion
NFT collections play a central role in the nowadays NFT ecosystem on all blockchains. It gives the ability to artists to organize their artworks and present them in a more meaningful way. It also can facilitate the use of other NFT features such as traits. As we'll present in this proposal, it can also enable more advanced features such as artist collaborations. These advanced features can get arbitrarily complex with the use of smart contracts.
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+            "power": 100 # this means the maximum possible power is 100
+        },
+        {
+            "speed": 180 # this means the maximum possible power is 180
+        }
+    ],
+    "stats": [
+        {
+            "rank": 50 # this means the maximum possible rank is 50
+        }
```
HazeyOneKenobi:
```suggestion
        },
```

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+        {
+            "number of swords": 1000 # this means the maximum possible number of swords is 1000
+        }
+    ]
+}
+```
+It must be noted that the actual value for traits will be present in the NFT data (refer to [artwork standard]()) and the above numbers for levels and stats only specify the maximum range.
```
HazeyOneKenobi: Might be helpful to add example NFT metadata matching the example given? 

```json
{
	"properties": [
		{
			"sex": "male"
		},
		{
			"hair color": "brown"
		},
		{
			"eye color": "green"
		}
	],
	"levels": [
		{
			"power": 50
		},
		{
			"speed": 90
		}
	],
	"stats": [
		{
			"rank": 11
		},
		{
			"number of swords": 948
		}
	]
}

**Review by HazeyOneKenobi — COMMENTED** [review]
First review ever, sorry if I did it wrong kek

File: eip-0034.md [hunk 14]
```diff
@@ -0,0 +1,121 @@
...
+
+## Design
+We consider collection NFTs as tokens! The issued collection token will contain all the collection information.
+
+An NFT will be considered as part of an already created collection if it follows the [artwork standard]() and the collection NFT is present in the first input of the issuance transaction which is known as the royalty box.
+
+There is no limitation on the number of issued tokens for a collection. One possible interpretation of such a feature is that artists can come together and create a collection. Each one of them will have one collection token which gives the right to them to add NFTs to that collection. We believe that with the use of smart contracts, more advanced usage of this flexibility will be discovered in the future.
```
nirvanush, line 17: Some pre-defined limitations protected in SC would be a cool feature actually. 
collection NFT mint power expiration date - can't mint more NFT for that collection after a specific date

eip-0034.md
```diff
@@ -0,0 +1,121 @@
...
+{
+    "name": "NAME_OF_THE_COLLECTION",
+    "description": "DESCRIPTION_OF_THE_COLLECTION",
+    "logo": "URL_TO_COLLECTION_LOGO",
+    "featured_image": "URL_TO_COLLECTION_FEATURED_IMAGE",
+    "banner_image": "URL_TO_COLLECTION_BANNER_IMAGE",
+    "category": "COLLECTION_CATEGORY",
+    "explicit_content": true or false
+}
+```
```
nirvanush: Not sure if should be a collection parameter or NFT's, but copyright rights can be interesting to have. 
NFT artist should define what the owner is allowed to use the NFT for: commercial use, modifications,

nirvanush (approved):

babygrenade:
Using numerator/denominator as a standard for stats values seems like it might be more flexible.  It could be used to represent non decimal numeric values.  Something like:

"stats": [
        {
                "name": "score",
                "numerator": "80",
                "denominator": "85"
        }

eip-0034.md
anon-real: This is in fact optional -- it can be an empty JSON. There is no limitation on what the keys can be, so the artist can also share his website. The list won't work because of the difficulties that it causes for UI/UX for marketplaces.

eip-0034.md
anon-real: Can you elaborate, please?

anon-real:
> Would there be something similar to Cardano's policy id?

The collection token will have a unique ID. I don't see any use-case for an additional off-chain ID. How would that be useful?

eip-0024.md
anon-real: Please let me know if it is better now.

eip-0034.md
anon-real: This part has changed altogether. Now the traits are only present in the NFT metadata and not in the collection. After some back and forth, this seems a more reasonable design.

eip-0034.md
anon-real: This part has changed altogether. Now the traits are only present in the NFT metadata and not in the collection. After some back and forth, this seems a more reasonable design. Please let me know if this is resolved.

eip-0034.md
anon-real: This part has changed altogether. Now the traits are only present in the NFT metadata and not in the collection. After some back and forth, this seems a more reasonable design. Please let me know if this is resolved.

eip-0034.md
anon-real: Protection is SC can not happen as tokens can not be necessarily tied to SCs. But having expiry data was part of the design which is not pushed yet. I'll push that soon as well.
This expiry date can be enforced by marketplaces.

eip-0034.md
anon-real: > Using numerator/denominator as a standard for stats values seems like it might be more flexible. It could be used to represent non decimal numeric values. Something like:
> 
> "stats": [ { "name": "score", "numerator": "80", "denominator": "85" }

I believe this is an unnecessary complication. I believe marketplaces like OpenSea don't have such a feature as well.

eip-0024.md
```diff
@@ -7,24 +7,125 @@
...
+            "value": 30,
+            "max": 80,
+        }
+    ]
+}
+```
+
+- If the artwork is part of a collection, R8 of this box contains token ID of the collection as Coll[Byte] otherwise it is an empty Coll[Byte]. This is needed for two cases:
+  - if the issuer box contains more than one collection token, this will specify which one is the one that the artist intended to put the artwork in.
+  - if the issue box contains some collection tokens but the artist does not intend to put the artwork in any collections. In this case, this register must be an empty Coll[Byte].
```
aslesarenko: Please describe the case when R8 is not defined.

eip-0024.md
```diff
@@ -7,24 +7,125 @@
...
+In this version of the standard, unlike V1, we don't specify how the royalty will be sent to the receivers. In V1, the proposition bytes of the issuer box receive the royalty amount. However, in this version, the royalty amounts may go to receivers specified in R6 or the proposition bytes of the issuer box depending on the [auction contract](eip-0022.md) standard. So NFT developers need to be careful about both R6 and the proposition bytes of the issuer box to be compatible.
+- R7 of this box may contain a JSON specifying trait values for the NFT as Coll[Byte]. 
+There are three types of traits:
+  - Properties: These are textual traits such as specifying `sex` as `male`.
+  - Levels: These are numerical traits that encode some sort of level-like information such as specifying `speed` as `60 out of 100`.
+  - Stats: These are numerical traits that encode any numerical information about the NFT such as specifying `age` as `25 out of 50`.
```
aslesarenko, line 35: Using Json have two major drawbacks. 
1) it is not compact, a lot of space is wasted by storing repeated "key", "value", "max" names, which are not really necessary.
2) The content cannot be accessed from smart contracts (this may be not necessary now, but no doubt will find it's use case in future)

I suggest to use the following Sigma type, which is isomorphic to the json below.

```
( // properties
  Coll[(  
    Coll[Byte],   // key
    Coll[Byte]    // value
  )],
  (  // levels
     Coll[(
        Coll[Byte], // key
        (Int, Int)     // value, max
      )],
     // stats
     Coll[(
        Coll[Byte], // key
        (Int, Int)     // value, max
      )]
  )
)
```

A library of generic encoders/decoders can be implemented in Appkit (and in JS if necessary) to implement such canonical conversions Json <-> ErgoValue and avoid doing it in all dApps again and again.

aslesarenko (changes requested):

eip-0034.md
```diff
@@ -0,0 +1,62 @@
...
+## Unlockable Content
+Unlockable content information must be present in `R7` of the issuance box as `Coll[Byte]`.
+Unlockable contents are files or other types of information that the artist reveals to the buyer when NFT is sold (and/or some other conditions are satisfied). At the time of proposing this EIP, there are no known decentralized solutions to have unlockable contents. However, it is more than likely that we'll have solutions for this soon so this part of this EIP introduces a structure that can be changed for backward and forward compatibility in the future.
+The following JSON contains the structure for unlockable contents.
+
+```json
+{
+    "version" standard_version, # this will change as new standards are needed to be added for unlockable contents
+    "type": "UNLOCKABLE_CONTENT_TYPE", # this will contain the type of unlockable content such as image
+    "content": "CONTENT" # this will include the content to be unlocked in a way such that it is not revealed until some conditions are met -- this can be a URL or some other thing depending on "type"
```
aslesarenko: Could you elaborate on the revealing protocol?

eip-0034.md
```diff
@@ -0,0 +1,62 @@
...
+```json
+{
+    "name": "NAME_OF_THE_COLLECTION",
+    "description": "DESCRIPTION_OF_THE_COLLECTION",
+    "logo": "URL_TO_COLLECTION_LOGO",
+    "featured_image": "URL_TO_COLLECTION_FEATURED_IMAGE",
+    "banner_image": "URL_TO_COLLECTION_BANNER_IMAGE",
+    "category": "COLLECTION_CATEGORY",
+    "explicit_content": true or false
+}
```
aslesarenko: I think it is better to store this using ErgoValue encoding, rather than json. See my another comment above.

aslesarenko (changes requested):

eip-0024.md
mgpai22: This is a great idea!

anon-real: @aslesarenko It's actually a great idea.

anon-real: @aslesarenko how would Json <-> ErgoValue work for converting ErgoValue to Json if we have no keys in ErgoValue?

aslesarenko: @anon-real If you are talking about names of Json fields, then they are should be known from the context (basically they are static metadata, in my example above I put the names in comments).
If something is not static, then it is data and should be stored in the data structure itself.

aslesarenko (approved):

kii-dot (approved):

eip-0034.md
ldgaetano: So to clarify, the traits section will be in the issuer box or the issuance box?
Some of us are looking to use GetBlok.io's Plasma library for manipulating AVL trees in order to scale the NFT minting/selling process to 10k/100k/100M+ NFTs. Unfortunately, the use of traits in the issuer box that differ from NFT to NFT complicates this process and would seem to make more sense if it were part of the NFT issuance box itself. The rational is that traits are intuitively associated to the actual existing object (i.e. the NFT itself/issuance box) and not the precursor object that gave birth to it (i.e. the issuer box).

ldgaetano: Furthermore, to clarify again as well, the traits are only max/min values or specific values? If they are max/min values, then they would be the same for all NFTs in a collection or all NFTs with the same rarity for example?

ldgaetano: The problem encountered with this implementation of traits and using Plasma is that the NFT meta-data to be placed in the issuance box is stored in the avl tree. If issuer boxes are the same for all NFTs in a collection or NFTs of the same "rarity" then one could reuse these contracts and issuer boxes in subsequent minting txs, since using Plasma enables the possibility of "minting-on-demand" at the time of the sale. Similarly, it is the same with a random NFT sale when NFT information is not known ahead of time and only at the time of the sale/minting tx. Thus the registers cannot be prefilled unless with a previous tx.

ErgoBass:
Hello, I am reviewing this and if seems like it is very confusing on what is being done here. I am comparing this to the Cardano Policy ID standard, which is defined in CIP-1855. I believe this is what needs to be considered as a starting point for collections on Ergo. Basically, the Policy ID should be script, that " includes hashes of keys needed to forge new tokens and must be witnessed by these keys in such a way as the script stipulates."

This Script should associate a wallet key to a specific Policy ID. That policy ID should only be able to be used if it is within the parameters of the script and is witnessed by the correct key. The script should allow a period of time or number of assets to be set, after which the policy locks and no other tokens can be minted from this policy. 

Please read through CIP-1855 for more information. 

https://github.com/cardano-foundation/CIPs/tree/master/CIP-1855

anon-real:
> Hello, I am reviewing this and if seems like it is very confusing on what is being done here. I am comparing this to the Cardano Policy ID standard, which is defined in CIP-1855. I believe this is what needs to be considered as a starting point for collections on Ergo. Basically, the Policy ID should be script, that " includes hashes of keys needed to forge new tokens and must be witnessed by these keys in such a way as the script stipulates."
> 
> This Script should associate a wallet key to a specific Policy ID. That policy ID should only be able to be used if it is within the parameters of the script and is witnessed by the correct key. The script should allow a period of time or number of assets to be set, after which the policy locks and no other tokens can be minted from this policy.
> 
> Please read through CIP-1855 for more information.
> 
> https://github.com/cardano-foundation/CIPs/tree/master/CIP-1855

@ErgoBass could you elaborate on what exactly you think is the issue with this EIP?

kushti (approved):

PR #71, opened 2022-06-08
eip4 - 3d Model Asset Type

KrasaviceBlasen:
Simple addition to eip4 to include 3d models.
Similar to audio files, 3d models may require a cover image as seen in R9.

### DIFF
--- a/eip-0004.md
+++ b/eip-0004.md
@@ -53,6 +53,8 @@ The standardization of various asset types can be found below:
 | NFT - picture artwork              | [0x01, 0x01] - i.e., "0e020101"                        | SHA256 hash of the picture    | Optional - link to the artwork (UTF-8 representation) |
 | NFT - audio artwork              | [0x01, 0x02] - i.e., "0e020102"                        | SHA256 hash of the audio    | Optional - link to the audio encoded as Coll[Byte] or (link to the audio, link to the image cover) encoded as (Coll[Byte], Coll[Byte]) (UTF-8 representation) |
 | NFT - video artwork              | [0x01, 0x03] - i.e., "0e020103"                        | SHA256 hash of the video    | Optional - link to the video (UTF-8 representation) |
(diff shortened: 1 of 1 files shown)


capt-nemo429 (approved):

MrStahlfelge:
Not sure if it was better to add a asset type for arbitrary file types so that we don't need to add new types for every file type. 
For example, we could use R7 0x01, 0xFF + (file extension in bytes)
Or R7 0x01, 0xFF and file extension in R9 (see below)

For R9, we could use [EIP-29 attachments](https://github.com/ergoplatform/eips/blob/136d53d4435dbd4d5049eccc1060e7c8ae849f77/eip-0029.md) with the multi attachment with attachment #0 file extension, attachment #1 http link and attachment #2 cover link image (optional)

KrasaviceBlasen:
> Not sure if it was better to add a asset type for arbitrary file types so that we don't need to add new types for every file type. For example, we could use R7 0x01, 0xFF + (file extension in bytes) Or R7 0x01, 0xFF and file extension in R9 (see below)

Yeah, I would argue that a solution of this kind is more ideal, however the addition I wrote here is trivial to implement and as I mentioned on Discord this is the only thing needed for Comet to release their 3d NFTs. So a timely, trivial solution is what I opted for.

If you aren't already working on this, I would be happy to help update eip4 to use eip29 (and engage with opinions from the community), but for now I don't see any issues with adding the new asset type since it is straightforward and shouldn't cause any new complications for future changes to eip4.

capt-nemo429:
I think this PR is fine as it is, as it follows the pattern of previous described assets. Seems like a natural path to follow.

MrStahlfelge:
I am not sure if it is enough to know that an asset is a 3d model as there are a lot of different types of 3d models.
Sure, obj is quite of a standard, but what if the 3d asset linked here is something else? When ipfs is used, we don't always have the file extension available

sb01101001:
The other asset types don't specify the file formats so I feel this is the same for them? 
I think most general applications will likely only use the cover image, but for applications intending to use the 3d model they usually support a wide variety of file formats anyway.

MrStahlfelge:
> The other asset types don't specify the file formats so I feel this is the same for them?

True, it is a shortcoming that we can't fix anymore but should not be the reason to proceed in the same way. :-)

PR #72, opened 2022-06-14
EIP4 - NFT File Attachment Asset Type

KrasaviceBlasen:
Adding NFT File Attachments to eip4 asset types.
File Attachments can be used to attach an arbitrary collection of files to an NFT.
Benefits of this eip:
- Every new file format can use NFT file attachments and circumvent the need to create a new pull request to eip4 just to add a asset type
- Artists can provide authentic copies of additional files to an NFT (such as a cover photo, or bonus files)

### DIFF
--- a/eip-0004.md
+++ b/eip-0004.md
@@ -67,7 +67,7 @@ An NFT File Attachment should have a multi-attachment content type as per the [E
 - Attachment 2 (Optional): `contentType = 2` (plain text) - Link to the second file (the sha256 hash of this file should be in index 1 of R8 and the file extension should be given as the second file extenstion in attachment 0)
 - Attachment *n* (Optional): `contentType = 2` (plain text) - Link to the *n*th file (the sha256 hash of this file should be in index `n-1` of R8 and the file extension should be given as the *n*th file extenstion in attachment 0)
 
-**NFT File Attachment Examples**
+**NFT File Attachment Example**
 
 Files to be attached:
 | File Format        | File Link                                     | File Hash                      |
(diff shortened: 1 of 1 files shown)


MrStahlfelge (approved):
Looks good to me

eip-0004.md
```diff
@@ -53,6 +53,34 @@ The standardization of various asset types can be found below:
 | NFT - picture artwork              | [0x01, 0x01] - i.e., "0e020101"                        | SHA256 hash of the picture    | Optional - link to the artwork (UTF-8 representation) |
 | NFT - audio artwork              | [0x01, 0x02] - i.e., "0e020102"                        | SHA256 hash of the audio    | Optional - link to the audio encoded as Coll[Byte] or (link to the audio, link to the image cover) encoded as (Coll[Byte], Coll[Byte]) (UTF-8 representation) |
 | NFT - video artwork              | [0x01, 0x03] - i.e., "0e020103"                        | SHA256 hash of the video    | Optional - link to the video (UTF-8 representation) |
+| NFT - File Attachments             | [0x01, 0xFF] - i.e., "0e0201FF"                        | Collection of SHA256 hashes of the files encoded as Coll[Coll[Byte]]    | NFT File Attachment (see encoding below) |
```
kushti: Why 0xFF and not more modest choice (e.g. 0x0F) ?

kushti (approved):

eip-0004.md
KrasaviceBlasen: File attachments should be an all-encompassing NFT asset type, so 0xFF seemed to make sense as it was categorically different to the current NFT asset types.

However, I think you are right and we should use something more modest and reserve 0xFF for any dramatic future types.

0x0F would make sense, what do you think @MrStahlfelge ?

MrStahlfelge: Sure. At the end of the day, it does not really matter what we use here. The idea was, as you said, to set it apart from the existing types. 0x99 would also work, as 0x0F.

anon-real (approved):
LGTM

PR #82, opened 2022-10-17
ErgoPay support multiple transactions and addresses

MrStahlfelge:
First enhancement for ErgoPay: support signing multiple transactions at once
Second enhancement for ErgoPay: support using multiple addresses

### DIFF
--- a/eip-0020.md
+++ b/eip-0020.md
@@ -215,9 +215,8 @@ We recommend to use an URL request if the payload exceeds 400 chars.
 Ergo Wallet App 1.6 or higher https://github.com/ergoplatform/ergo-wallet-app
 
 ## Implementation in dApp
-ErgoPay showcase example 
-
-* https://github.com/MrStahlfelge/ergopay-server-example
+* ErgoPay showcase example https://github.com/MrStahlfelge/ergopay-server-example
+* Ergoplatform Jackson Serialization Library https://github.com/MrStahlfelge/ergoplatform-jackson
 * TokenJay https://tokenjay.app
 * Mosaik tutorial https://docs.ergoplatform.com/dev/stack/mosaik/intro/
 * Spectrum DEX
--- a/eip-0020.md
+++ b/eip-0020.md
@@ -268,9 +268,9 @@ HTTP header to indicate support: `ErgoPay-MultipleTx: supported`
 If a wallet app sets this HTTP header field, the `ErgoPaySigningRequest` might contain a 
 `reducedTxList` array field instead of the `reducedTx` field. All other fields stay the same.
 
-The `ErgoPaySigningRequest` must not contain `reducedTx` and `reducedTxList` fields.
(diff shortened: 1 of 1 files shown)


eip-0020.md
```diff
@@ -156,7 +158,7 @@ The wallet application should request URL and obtain the following data (json fo
 
 ```
 ErgoPaySigningRequest:
-  - transaction: ReducedTransaction (optional*)
+  - reducedTx: ReducedTransaction (optional*)
```
aslesarenko, line 161: I don't think we can just change the schema. If this change is really needed, then the spec should be versioned. So the current version is v1 and the change should be in version v2.

eip-0020.md
```diff
@@ -244,3 +246,31 @@ Any wallet application team should consider supporting ErgoPay in their wallet a
...
+needs to sign multiple transactions at once it needs multiple steps. To make it more convenient to 
+sign multiple transactions, especially useful for chained transactions, the described ErgoPay 
+enhancements supports transferring multiple transactions at once.
+
+HTTP header to indicate support: `ErgoPay-MultipleTx: supported`
+
+If a wallet app sets this HTTP header field, the `ErgoPaySigningRequest` might contain a 
+`reducedTxList` array field instead of the `reducedTx` field. All other fields stay the same.
+
+The `ErgoPaySigningRequest` must not contain `reducedTx` and `reducedTxList` fields.
```
aslesarenko:
```suggestion
The `ErgoPaySigningRequest` must not contain `reducedTx` and `reducedTxList` fields simultaneously.
```

eip-0020.md
```diff
@@ -244,3 +246,31 @@ Any wallet application team should consider supporting ErgoPay in their wallet a
...
+
+Signed transactions are submitted to Ergo network in their order in the `reducedTxList` array.
+
+In case `reducedTxList` array field and `replyToUrl` field is set, the data posted to this url is an
```
aslesarenko:
```suggestion
In case `reducedTxList` array field and `replyToUrl` field is set, the data POSTed to this url is an
```

Please also specify format of the POST message with the transaction id strings.

aslesarenko (changes requested):

eip-0020.md
MrStahlfelge: True, this is no change but a documentation fix. I should have mentioned it. The field was named wrong here (or in the code) and the real name used in the wild is `reducedTx`: https://github.com/ergoplatform/ergo-wallet-app/blob/23454d820026805e8eeabc7433a3f6831c1f2a1d/common-jvm/src/main/java/org/ergoplatform/transactions/ErgoPay.kt#L79

aslesarenko (approved):

Luivatra:
Would be very useful for ergopad to be able to redeem multiple vestings at once

deadit (approved):

deadit (approved):

PR #87, opened 2022-12-17
EIP-41 Stealth address standard

ross-weir:


eip-0041.md
```diff
@@ -0,0 +1,148 @@
...
+- Register `R6`
+  - Type: `SConstant[SGroupElement]`
+  - Value: ur = u^r
+- Register `R7`
+  - Type: `SConstant[SGroupElement]`
+  - Value: ur = u^r
+
+> 📝 As discussed in the `ergoforum` discussion [[1]](#1) this register declaration is larger in size than that originally proposed by `kushti` but possesses the useful property that it could look like a legitimate use-case.
+
+## `Receiver`s public address and wallets
```
ross-weir: Maybe better ways of doing this but the main intent is:

1. Enable stealth boxes to be retrieved by `Receiver` after a fresh wallet restore
2. Introduce very minimal friction to `Sender` when making a transaction (ideally none - no change in the user journey when making a stealth payment)

arobsn: On point .1, it's an interesting way to make Stealth Addresses scalable since currently, we need to fetch all Stealth Boxes and test them one by one if they are spendable by a given secret key. 

However, I see two problems here: (1) an address is an encoded representation of an ErgoTree, so in order make to make it two-way reproducible (ErgoTree <--> Address) we need to add the public key in the ErgoTree body, in other words, the above ErgoTree will be encoded as `2pxnLshTr2aL5t3QFd6ScSEHC5Jk6EpnMRzGb3` by the node, no matter what public key is used. Which leads us to another problem (2): it's exposing a recipient's public key. We attenuate this problem if we use a public key that we'll have some level if trust that will not be used as a regular P2PK address through standard means, something like a different derivation `account` path for Stealth Addresses, `m/44'/429'/1'/0/0`, for example. But I'm not sure about real privacy gains here. From what I can see, it's not different than generating a new P2PK address to receive funds.

eip-0041.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. `Sender` uses a regular payment exposing a transaction that `Receiver` wanted to remain private
+2. Introduces friction when making payments, possiblily reducing conversions
+3. Less user friendly, `Sender` as a non-technical user might not know about stealth payments
+
+Ideally when `Sender` is creating a stealth transaction in their wallet the UX flow should remain the same as a regular transaction, the fact that the `Receiver` is requesting a stealth payment should be opaque to the `Sender` when making the payment. To achieve this we could encode a "payment type" in the public address so wallets can detect the `Receiver` is requesting a specific type of payment and construct the transaction accordingly, similar to network and address type prefixes but only for supporting wallets.
+
+This could extend in the future to more non-interactive custom payment types, not just stealth payments.
+
+**Encoding**
```
arobsn: Encoded Ergo addresses already have a head byte for network and address type encoding. Which is composed as `Network + AddressType`, as defined bellow:

```ts
enum Network {
    Mainnet = 0,
    Testnet = 16
}
```
and 

```ts
enum AddressType {
    P2PK = 1,
    P2SH = 2,
    P2S = 3
}
```

`AddressType` can be extended by adding `P2SA = 4` and the `Pay to Stealth Address` can be encoded as follows:

```ts
const headByte = Uint8Array.from([Network.Mainnet + AddressType.P2SA]);
const publicKey = hexToBytes("038d39af8c37583609ff51c6a577efe60684119da2fbd0d75f9c72372886a58a63");

const body = concat(headByte, publicKey)
const checksum = blake2b256(body).subarray(0, 4);

const encodedAddress = base58(concat(body, checksum));
```

The code above will result in `encodedAddress == "bSRib1gpEX5u3EEysRwgSsVLerpmPsLWDyrUNSoMv4t1Cc2a1vh"` for mainnet, for testnet it will result in `3xiWsAn5chBHVKSXq5kEREPHKc5G86Cy7HwwRdzKDa8GWVjfQgFt`. Addresses will be prefixed by `b` for mainnet and `3x` for testnet.

A wallet can check if the address is a Pay to Stealth Address by checking `headByte & 0xf == AddressType.P2SA`.

eip-0041.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+1000cee4c6a70407e4c6a70507e4c6a70607e4c6a70707
+```
+
+**Generation of stealth box registers [[5]](#5):**
+
+```typescript
+const g = new EC("secp256k1").g; // group element generator
+const u = receiverPublicAddress;
```
arobsn: I think we can avoid some confusion by replacing `receiverPublicAddress` by `receiverPublicKey`.

arobsn (changes requested):

eip-0041.md
ross-weir: I can't see a way of including the pubkey in the ErgoTree body while preserving receiver privacy (with my limited crypto experience) even with a different derivation path

For example:
1. Alice posts public key for receiving stealth payment
2. Bob makes payment to Alice including her pubkey in the ErgoTree so it can be easily looked up
3. Anyone can see that Bobs payment was to Alice because of the presence of the Alices pubkey she posted in the ErgoTree

eip-0041.md
ross-weir: My thinking behind this was that stealth addresses seemed like a layer on top of "core" address types (`AddressType` enum you described) since they boil down to a P2S address transaction

If we want to add `AddressType.P2SA` that would be cool, simpler and straight forward

arobsn: Yeah, I think this approach seems more appropriate for a standardized address encoding.

eip-0041.md
arobsn: Exactly, so might be worth removing this from the EIP scope and describing how to test if Stealth Boxes are spendable.

eip-0041.md
```diff
@@ -0,0 +1,139 @@
...
+
+```ts
+enum AddressType {
+    P2PK = 1,
+    P2SH = 2,
+    P2S = 3
+}
+```
+
+We propose adding a new address type of `AddressType.P2SA`:
```
arobsn, line 113: It would be good to describe how to encode a P2SA and make clear that it's a special kind of address, as it's not ErgoTree-based like others. 

You can take this comment as the basis for that: https://github.com/ergoplatform/eips/pull/87#discussion_r1051699979

arobsn (changes requested):

aslesarenko:
1) There is no need to use registers to store gr, gy... parameters. Storing in a register of SELF (especially when accessed unconditionally with .get ) is always equivalent to embedding those values in the contract body (i.e. in ErgoTree). This is not only functionally equivalent, but also stored in the same box, just in different register.
2) When registers are not used and parameters are embedded, no need to use special P2SA, because P2S will be enough and different receiverPublicKey will generate different P2S addresses.
3) this will also solve scalability problem
4) this encoding saves some space in box storage also
5) all information necessary for wallet and other off-chain tools can be extracted from p2s address alone.
6) when wallet user want to receive stealth payment, he need to choose the payment type, and then the wallet can generate p2s stealth address.
7) even when gr, gy... parameters embedded in ErgoTree, they still can be segregated into constants array of ErgoTree and serve as paremeters of contract template
8) talking about templates. wallets can adopt EIP-5 to generate parameterised addresses for different use cases. Stealth Address is one such use case, so it is a good application of Contract Template EIP-5

ross-weir:
> 1. There is no need to use registers to store gr, gy... parameters. Storing in a register of SELF (especially when accessed unconditionally with .get ) is always equivalent to embedding those values in the contract body (i.e. in ErgoTree). This is not only functionally equivalent, but also stored in the same box, just in different register.
> 2. When registers are not used and parameters are embedded, no need to use special P2SA, because P2S will be enough and different receiverPublicKey will generate different P2S addresses.
> 3. this will also solve scalability problem
> 4. this encoding saves some space in box storage also
> 5. all information necessary for wallet and other off-chain tools can be extracted from p2s address alone.
> 6. when wallet user want to receive stealth payment, he need to choose the payment type, and then the wallet can generate p2s stealth address.
> 7. even when gr, gy... parameters embedded in ErgoTree, they still can be segregated into constants array of ErgoTree and serve as paremeters of contract template
> 8. talking about templates. wallets can adopt EIP-5 to generate parameterised addresses for different use cases. Stealth Address is one such use case, so it is a good application of Contract Template EIP-5

Thanks a lot for the feedback!

Embedding values is something I thought about but decided against because the parameters (gr, gy...) are created from one-time secrets generated by the sender. Because there's no way for the receiver to know what values the sender used there's no way to know if a P2S address is a stealth payment for them, they'd still need to check all P2S addresses that look like a stealth payment to see if they're spendable - they may even check P2S addresses that look like stealth payments but aren't

PR #88, opened 2023-01-21
EIP-0042 Multi-Signature Wallet

vorujack:


eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+{
+   MSR: "{\"tx\": ...}",
+   n: <number of chunks>
+   p: <page number>
+}
+```
+
+This data is transferred between multi-signature wallets until one wallet has at least `N` commitments (including its own).
+
+This wallet creates a partially signed transaction with this encoding:
```
MrStahlfelge, line 146: Why does the last wallet create a "partially signed transaction" and not a fully signed transaction if it already has N commitments? Or does this sentence apply until N-1 commitments are collected?

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+
+```
+{
+    partialTx: "<partially signed transaction encoded as base64>",
+    commitments: <ancoded commitments>
+    signed: [<base64 encoded of signed public keys>],
+    simulated: [<base64 encoded of simulated public keys>],
```
MrStahlfelge, line 153: What means simulated? This is only the list of missing keys, which can be determined by the wallet address. Why do we need it?

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+
+This data is transferred between multi-signature wallets until one wallet has at least `N` commitments (including its own).
+
+This wallet creates a partially signed transaction with this encoding:
+
+```
+{
+    partialTx: "<partially signed transaction encoded as base64>",
+    commitments: <ancoded commitments>
+    signed: [<base64 encoded of signed public keys>],
```
MrStahlfelge: Can you give actual examples here? A public key can be stored in many different encodings, as ergotree, as an ergo address etc. It is not clear what format is used here.

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+]
+```
+
+In the multi-signature wallet, a commitment is represented as a list of base64 strings to reduce the commitment size. Also, an empty string is used wherever a commitment is unavailable.
+
+[//]: # (and we can transform these commitments to standard format:)
+This reduced format can be transformed into the standard format:
+
+```
+{
```
MrStahlfelge, line 120: This format differs from the actual format in [EIP-11](https://github.com/ergoplatform/eips/blob/9906f403f61d8592ab955494cc4310563f4067c6/eip-0011.md?plain=1#L104) (index/position, pubkey/pk, a/commitment). Is EIP-11 outdated?

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+  boxes:[encoded boxes as base64],
+  commitment: commitments encoded
+}
+```
+
+The input `boxes:` are only used to display transactions on wallets.
+
+The general commitment format is as follows. However, this general-purpose commitment can be used for any contract.
+```
+{
```
MrStahlfelge, line 93: Differs from EIP-11, see below

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+After an unsigned transaction is generated, the wallet
+
+Generates a commitment.
+Stores its private part locally.
+Shares a commitment as below
+
+```
+{
+  tx:<reduced transaction bytes encoded as base64>,
+  boxes:[encoded boxes as base64],
```
aslesarenko, line 84: Why base64? We consistently use base16 for binary data (byte arrays) across EIPs and API messages.

eip-0042.md
aslesarenko: Agree, if EIP-11 is not suitable "as-is", then please explain why and if motivated, then EIP-11 should be updated. Then this EIP should follow EIP-11.

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+    "a": "0253db866791af521ba4ab009509b6db89d272d9461636ee65eaa3e316884b21a4"
+}
+```
+
+In a multi-signature wallet, a simplified version of commitment is used.
+
+```
+commitments: [
+  [<base64 string of commitment for public key index 0 in input 0>, ...]
+  [<base64 string of commitment for public key index 0 in input 1>, ...]
```
aslesarenko, line 110: See my comment above about base64.

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+- The first byte in the array is `10`.
+- The number of constant values in the contract is encoded as a variable-length quantity (VLQ).
+  This number equals `M+1` (Constant `N` and `M` public keys)
+- A byte with value `04` denotes the integer type,followed by the value of `N` encoded as VLQ.
+- For each public key 35 bytes are used: 2 bytes with value `08cd`, followed by 33 bytes representing the public key.
+- A byte with the value `98` indicates **"atLeast"**.
+- Two bytes, with values `73` and `00`, respectively.
+  The value `00` indicates the VLQ-encoded index of value `N` in the list of constants.
+- A constant byte with value `83`, the VLQ-encoded value of `M`, and a constant byte with value `08`
+- For each public key (`M` times), a byte with value `73`,followed by the VLQ-encoded value of the index of the public key in the list of constants
```
aslesarenko: I suggest to use [ErgoTree spec](https://ergoplatform.org/docs/ErgoTree.pdf) and specify here the ErgoTree using operation names from the specification.
And and most magic bytes in the text above are actually opcodes, so it is better to mention them (like **"atLeast"**)

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+We must guarantee that all signers will have a unique and reproducible multi-signature wallet address. To do so, we use the same derivation path and sort the byte array of public keys in ascending order.
+
+Then we will manually create the ErgoTree; since using the compiler could cause the resulting address to vary.
+
+The address derivation algorithm is as follows:
+
+Sort the list of signers' public keys before being used in the contract.
+Create the required ergoTree array as follows:
+
+- The first byte in the array is `10`.
```
aslesarenko: The first byte is the header, which contains [ErgoTree version bits](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c4b9c90d4d01cff7022f4ca3f582836639e82f1a/sigmastate/src/main/scala/sigmastate/Values.scala#L1217). 
While it is ok to always use one specific version, in the EIP it is better to have `version` as parameter. 
And, if the `version > 0`, then `size bit 3` should be set, and the size of tree bytes should be stored in as necessary (see  section 5.5 in the spec)

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+- For each public key 35 bytes are used: 2 bytes with value `08cd`, followed by 33 bytes representing the public key.
+- A byte with the value `98` indicates **"atLeast"**.
+- Two bytes, with values `73` and `00`, respectively.
+  The value `00` indicates the VLQ-encoded index of value `N` in the list of constants.
+- A constant byte with value `83`, the VLQ-encoded value of `M`, and a constant byte with value `08`
+- For each public key (`M` times), a byte with value `73`,followed by the VLQ-encoded value of the index of the public key in the list of constants
+
+Therefore, for each derivation path, all wallets are able to generate a unique and similar byte array as an ErgoTree.
+
+![Ergo Tree Bytes](./eip-0042/bytes.png)
```
aslesarenko, line 69: Another problem with this specification, it is rather rigid. For example the constants not necessary at all and they all can be embedded in the AtLeast directly as arguments. This is basically the same contract, but more compact, saving at least 2 + 2*M bytes. It is, of course, will have different address.
In this regard I suggest make this optimization in this EIP.

eip-0042.md
```diff
@@ -0,0 +1,174 @@
...
+## Introduction
+
+On top of EIP-11, this EIP implements the following additional operations:
+
+* Wallet creation and address derivation.
+* Transaction signing based on EIP-11.
+
+## Wallet Creation and Address Derivation
+
+These parameters define an `N-out-of-M` multi-signature wallet:
```
aslesarenko: threshold is usually called k-out-of-n. Intuitively k is smaller than n. In this text it is reversed, so I found it quite confusing on the first read.

aslesarenko (changes requested):

eip-0042.md
mhssamadani: see https://github.com/ergoplatform/eips/blob/9906f403f61d8592ab955494cc4310563f4067c6/eip-0011.md#signing-procedure. 
It's a two-round protocol in which the parties generate commitments in the first round. The last party is able to partially sign (he has only one of the keys so he cannot fully sign) and then, each of them signs the transaction partially and sends the generated signature to another party. This process continues until the signature is completed.

eip-0042.md
mhssamadani: see https://storage.googleapis.com/ergo-cms-media/docs/ErgoScript.pdf, Appendix A2. Proving.
Those signers that do not participate in signing will be simulated. See above for details.

eip-0042.md
mhssamadani: It's 33-Byte base64 encoded public key. So no prefixes, suffixes, or any additional bytes.

eip-0042.md
mhssamadani: this format is used in the node, sigma-rust and other libs, see /wallet/generateCommitments.
I'm not sure why EIP-11 is not following this format.

eip-0042.md
mhssamadani: base64 results in smaller QR codes.

eip-0042.md
mhssamadani: Nice idea! will do.

eip-0042.md
mhssamadani: correct.

eip-0042.md
aslesarenko: @mhssamadani, is it the binary format used in GroupElementSerializer of Sigma?

vorujack: yes.

eip-0042.md
vorujack: Its updated.

eip-0042.md
vorujack: fixed

eip-0042/bytes.png
aslesarenko, line 1: There is a mistake. 
The "Collection Element Data Type (SigmaProp)" is not saved.
Please check ConcreteCollectionSerializer and ValueSerializer.

aslesarenko (changes requested):

eip-0042/bytes.png
vorujack: I'm not sure if I got it, please elaborate.

aslesarenko: On your picture, the "Collection Element Data Type (SigmaProp)" shown under "Repeated for each public key". This is not correct. That 08 value is stored only once. And you already showing it as "Collection Data Type (SigmaProp)"

vorujack: according to the doc, it seems you're right since Expression Serializations requires opCode and body; however, in the tests I'm getting parsing error if I eliminate those types and surprisingly it's working when type is added.

PR #91, opened 2023-03-06
EIP-43: Reduced Transaction

aslesarenko:
The original text was part of EIP-19. In this PR I moved it into separate EIP. 
Reduced Transactions are now used in ErgoPay and also make sense in other use cases (i.e. API endpoint)

- [x] moved background and motivation section
- [x] moved data structure section
- [x] Added specification of binary format.

### DIFF
--- a/eip-0043.md
+++ b/eip-0043.md
@@ -105,6 +105,7 @@ names are not serialized.
 | `messageSize`   | `VLQ(UInt)`         | Number of bytes in the messageToSign of the transaction |
 | `messageToSign` | `Bytes`             | serialized bytes of the unsignedTx.messageToSign  |
 | `reducedInputs` | `ReductionResult*` | serialized reduced inputs                         |
+| `txCost`        | `VLQ(UInt)`        | transaction cost according to the prover          |
 
 Note, the number of reduced inputs is serialized as part of the `messageToSign`.
 
--- a/eip-0043.md
+++ b/eip-0043.md
@@ -1,0 +1,129 @@
+# EIP-43: Reduced Transaction 
+
+* Author: @aslesarenko
+* Status: Proposed
+* Created: 18-August-2021
+* License: CC0
+* Forking: not needed
+
+## Contents
+- [Description](#description)
+- [Background And Motivation](#background-and-motivation)
+- [Reduced Transaction data](#reduced-transaction-data)
+- [Data Formats](#data-formats)
+- [Reference implementation](#implementation-in-appkit)
+
+## Description
+This EIP defines a data format for Reduced Transaction which can be used:
+- to enable interoperation between an online dApp and a wallet app for creating, signing and sending Ergo transactions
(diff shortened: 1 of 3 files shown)


eip-0043.md
```diff
@@ -0,0 +1,129 @@
...
+### ReducedTransaction
+
+ReducedTransaction is serialized to bytes array in the format described below. Note, field
+names are not serialized.
+
+| Field Name      | Format              | Description                                       |
+|-----------------|---------------------|---------------------------------------------------|
+| `messageSize`   | `VLQ(UInt)`         | Number of bytes in the messageToSign of the transaction |
+| `messageToSign` | `Bytes`             | serialized bytes of the unsignedTx.messageToSign  |
+| `reducedInputs` | `ReductionResult*` | serialized reduced inputs                         |
```
greenhat, line 107:
```suggestion
| `reducedInputs` | `ReductionResult*` | serialized reduced inputs                         |
| `txCost`        | `VLQ(UInt)`        | transaction cost according to the prover          |
```

greenhat: We have the tx cost also serialized as part of ReducedTransaction in sigma-rust since the early days.

greenhat (changes requested):
Looking good. Please check my comment on missing txCost.

eip-0043.md
aslesarenko: @greenhat, just checked in the git history, there has never been txCost in Appkit's reference implementation. It doesn't make sense to store it, because there are costs for ReducedResults which can be summed up.
There is probably a bug in sigma-rust. Please check how it is used in practice and does it really make sense to store it.

greenhat: Thats weird. I added it after our discussion here - https://discord.com/channels/668903786361651200/860527298292613151/912658058973437993

aslesarenko: @MrStahlfelge, I think the resolution largely depends on how this is implemented in ErgoPay and Mobile wallet, we cannot break all the dapps. Does it have txCost at the end of ReducedTransaction?

MrStahlfelge: Yes, Appkit saves the cost at the end:

https://github.com/ergoplatform/ergo-appkit/blob/5f15c28ad61ccef9dcd9cf6ab1407dcaa20539b2/lib-impl/src/main/java/org/ergoplatform/appkit/impl/ReducedTransactionImpl.java#L74

https://github.com/ergoplatform/ergo-appkit/blob/2bb32fb0db30381266865b19f96148dd5537744b/lib-impl/src/main/java/org/ergoplatform/appkit/impl/BlockchainContextBase.java#L49

aslesarenko: Ok, I moved cost saving to the serializer here https://github.com/ergoplatform/ergo-appkit/pull/217

greenhat (approved):

MrStahlfelge (approved):

PR #94, opened 2023-06-28
EIP-0023. Let ballot token owner (R4 in ballot box) always spend the ballot box

greenhat:
Following https://github.com/ergoplatform/eips/pull/41/files#r1243427074

## Motivation

`isSimpleCopy` clause for `owner` requires minting new ballot tokens for ANY change to the update contract or its parameters (min_votes). 

I forgot to mint new ballot tokens on changing the update contract and now those ballot tokens cannot be used. See https://github.com/ergoplatform/oracle-core/issues/291#issuecomment-1609089590 for details.

Rewriting the final clause as `owner || (isSimpleCopy && update)` would allow to keep using existing ballot tokens since voting is done by ballot owners (PK check).

### DIFF
--- a/eip-0023/contracts/ballot_contract.es
+++ b/eip-0023/contracts/ballot_contract.es
@@ -28,6 +28,5 @@
   
   val owner = proveDlog(selfPubKey)
   
-  // unlike in collection, here we don't require spender to be one of the ballot token holders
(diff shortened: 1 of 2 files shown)


scalahub:
Approved

PR #99, opened 2024-02-24
EIP-0045 Ephemereal Tokens

andrehafner:
Ephemeral Tokens Proposal

### DIFF
--- a/eip-0045.md
+++ b/eip-0045.md
@@ -1,0 +1,42 @@
+# Ephemeral Tokens
+
+- Author: qx()
+- Status: Proposed
+- Created: 23-February-2024
+- Last edited: 23-February-2024
+- License: CC0
+- Forking: not required
+
+## Abstract
+
+This document proposes a standard for creating and burning ephemeral tokens based on a future block height with incentive for miners to burn the token.
+Incentivized Burning!
+
+## Motivation
+
+There exists many use cases for ephemeral tokens, here are a few:
+- Blockchain messaging (messages that don't need to take up storage for longer than a specified time) including encrypted messaging (encryption is up the the clients of course)
+- Key access - tokens that grant rights that are then automatically burnt after a period of time (think of committee access)
+- Real-Time data - when data required is ONLY what is current (some sort of oracle perhaps), the old data can be pruned to save space on the chain
+- NFT collections - Rareness could be ranked by the point at which it would be expired, with the most valuable ones being the random ones that have a block height many years out, or the reverse.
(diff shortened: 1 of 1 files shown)


TheStophe:
First thing I had to do is look up the definition of Ephemeral: "The term "ephemeral" comes from the Greek word "ephemeros," meaning lasting only one day or short-lived. It describes things that exist or are used for a very brief period before disappearing or being forgotten. This concept applies broadly, from natural phenomena like morning dew, which vanishes as the day warms, to digital technology where ephemeral data or messages are designed to disappear after a certain time to ensure privacy or manage storage efficiently."

TheStophe:
Quick Summary of EIP-0045:

- Proposes a standard for creating and burning ephemeral tokens at a specific future block height, with incentives for miners.
- Aims to serve use cases like blockchain messaging, temporary access tokens, real-time data management, and expiring NFT collections.
- Requires an ephemeral smart contract (SC) that outlines token creation, transfer, and burning processes.
Incentivizes miners to burn tokens post-specified block height by allowing them to collect ERGs from the smart contract.
- Involves an off-chain bot for minting tokens according to SC requirements and facilitates ownership transfer and miner incentives.

TheStophe:
Seems like a lot of good use cases and logic for this type of token SC, great idea. 

I would just like to understand the technical side and risks in a little more detail and from the collective intelligence here.

So from the users perspective, I send Erg to the Ephemeral SC addr, then I get the EIP-45 token in my wallet. When the block height is reached, a miner will consume the token, and it will automatically disappear from the wallet?

andrehafner:
> Seems like a lot of good use cases and logic for this type of token SC, great idea.
> 
> I would just like to understand the technical side and risks in a little more detail and from the collective intelligence here.
> 
> So from the users perspective, I send Erg to the Ephemeral SC addr, then I get the EIP-45 token in my wallet. When the block height is reached, a miner will consume the token, and it will automatically disappear from the wallet?

Tokens never are in your wallet, always in sc. you just have permission to transfer them to someone else’s wallet ownership and they have your wallet address in there indicating that your wallet is the owner (for other dapps to verify).  Otherwise there would be no way to force burning as someone could hold token in wallet with .001 ergs in box (no miner incentive there). You send erg to a proxy contract and then an off chain bot mints it for you in the sc. similar how your reward GORTs live in the contract and sit there until you extract (but these you can’t extract).

TheStophe:
Ah thank you. A diagram of how it might work lifecycle would be visually helpful, but you explained it well. Thanks.

PR #100, opened 2024-04-09
[WIP] EIP-50 - 6.0 Soft-Fork

kushti:


eip-0050.md
```diff
@@ -0,0 +1,112 @@
...
+issue: https://github.com/ScorexFoundation/sigmastate-interpreter/issues/554
+
+
+* finish executeFromVar implementation
+
+issue: https://github.com/ScorexFoundation/sigmastate-interpreter/issues/443 , https://github.com/ScorexFoundation/sigmastate-interpreter/issues/612
+
+* support MethodCall encoding of Numeric methods
+
+issue: https://github.com/ScorexFoundation/sigmastate-interpreter/issues/667
```
aslesarenko, line 93: Maybe this is not necessary. Looking back is was not good idea to allow both `method call` and `opcode` encoding for methods. It would be better to forbid method calls for those methods which have dedicated opcodes.
The translation MC -> opcode is actually happens in the compiler, but is not strictly enforced. This creates more surface for test coverage and potential bugs.

eip-0050.md
```diff
@@ -0,0 +1,112 @@
...
+
+Tests:
+
+
+* Add compilation test in *TestingInterpreterSpecification*
+
+* Add roundtrip test in *MethodCallSerializerSpecification*
+
+* Add evaluation test (see "nbits evaluation" test)
+
```
aslesarenko, line 199:
```suggestion
* For each new method separate property tests should be added to SigmaDslSpecification.
```

aslesarenko (dismissed):

eip-0050.md
kushti: I think it would be better to develop ErgoTree via adding methods, to have simple universal description of the language.

eip-0050.md
kushti: appendix removed now

kushti:
@aslesarenko comments addressed , please make another pass

PR #101, opened 2024-06-27
EIP-0046: Atomic Transaction Chains

aslesarenko:


anon-real (commented):
This feature would be great! There has been a few instances in my dApps that I needed this. To increase the probability of txs being mined together, I set the fee for all txs except the last one to the minimum possible ERG amount.

Expanding on that, one other way to achieve this effect would be to:
- have an api for batch submission of txs (instead of one by one)
- in batch submission, allow txs with zero fees, as long as the zero-fee tx is a parent of a non-zero fee tx.

With this approach, there is no incentive for miners to mine zero-fee txs unless the average fee per byte of the whole batch is meaningful for them (so they will have to include all txs to get the fee)

PR #102, opened 2024-07-08
EIP-0047: Pooled Transaction Inputs

aslesarenko:


Luivatra:
- Will need to add token support as well, besides just the value
- Besides leaving the specific inputs undefined at time of signing, would it be possible to leave some outputs undefined as well? Example:
  - Inputs: 
    - Some spectrum lp pooled input (spending 1000 sigusd) 
    - user pooled input (spending 10 erg)
  - Outputs: 
    - Some spectrum lp output (~10 erg more than input)  
    - miner box 
    - user change box (1000 sigusd + leftover erg)

aslesarenko:
> * would it be possible to leave some outputs undefined as well?

I think it will be possible to add some flexibility to how Change Outputs are created. 
For example each PooledInput can have `sendChangeTo: Option[SigmaProp] = None` field, which is None by default.
When it is None then change is always returned to the pool, as described in the EIP.
However, when the change address is explicitly specified, then all the change can go to that address.

With this addition your above example will have 

- Inputs:
  - PooledInput(spectrumAddress, SigUSD: 1000, sendChangeTo: Some(userAddress))
  - PooledInput(userAddress, ERG: 10)
- Outputs:
  - Some spectrum lp output (~10 erg more than input)
  - miner box
 
So, no need to create change output explicitly, but still need to explicitly specify where the change should go from the pool.

